PR #25, opened 2019-09-25
pagination by Yevhen Sahach

eugeneSahach:
https://yevhen-sahach.github.io/react_pagination/


irenhh:
it looks like it should look
or what did you mean by that?

irenhh:
fix the error with key attribute in Pagination component

PR #26, opened 2019-09-27
solution

youvovk:
https://youvovk.github.io/react_pagination/

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,22 +1,41 @@
 import React from 'react';
+
+import Pagination from './Pagination';
+
 import './App.css';
 
 class App extends React.Component {
   state = {
-    tabs: [
-      { title: 'Tab 1', content: 'Some text 1' },
-      { title: 'Tab 2', content: 'Some text 2' },
-      { title: 'Tab 3', content: 'Some text 3' },
-    ],
+    page: 0,
+    perPage: 5,
+    total: 42,
   };
 
+  onPageChange = (changedPage) => {
+    this.setState({
+      page: changedPage,
+    });
+  }
+
+  onPerPageChange = ({ target: { value } }) => {
+    this.setState({
+      perPage: +value,
+    });
+  }
+
   render() {
-    const { tabs } = this.state;
+    const { page, perPage, total } = this.state;
 
     return (
-      <div className="App">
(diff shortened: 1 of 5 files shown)


src/App.js
```diff
@@ -1,22 +1,42 @@
...
 
+  onPageChange = (changedPage) => {
+    this.setState({
+      page: changedPage,
+    });
+  }
+
+  onPerPageChange = ({ target: { value } }) => {
+    this.setState({
+      perPage: +value,
```
irenhh: don't mutate the value

src/Pagination.js
```diff
@@ -0,0 +1,313 @@
...
+          additionalInfo: `${(index * perPage)}-${((index + 1) * perPage) > total
+            ? total
+            : ((index + 1) * perPage)}`,
+          active: false,
+        };
+      }),
+    });
+  }
+
+  isUnDisable = (page) => {
```
irenhh: `isUndisabled` or better `isEnabled`

irenhh: btw, what `isEnabled`?

irenhh: what are you gonna enable?

src/Pagination.js
```diff
@@ -0,0 +1,313 @@
...
+    const { onPageChange, preparedArrPages } = this.props;
+
+    onPageChange(page);
```
irenhh: you can't have two setstate inside one method, no need to create this func

src/Pagination.js
```diff
@@ -0,0 +1,313 @@
...
+      });
+    }
+  }
+
+  nextPage = (page) => {
+    const { onPageChange, preparedArrPages } = this.props;
+
+    if (page === preparedArrPages.length - 1) {
+      onPageChange(page);
+      this.setState({
```
irenhh: create one setState func and add conditions in it

src/Pagination.js
youvovk: Renamed method. I need this because Pagination it's controlled and I need to know  when page changed and enable my previous and next button

src/Pagination.js
youvovk: I need to follow on change page

PR #34, opened 2020-02-19
Pagination

AlexanderShpilka:
https://alexandershpilka.github.io/react_pagination/

Выполненные пункты: 1 - 5


src/App.js
```diff
@@ -1,6 +1,43 @@
...
+import { Pagination } from './components/Pagination/Pagination';
+import { PerPageSelector } from './components/PerPageSelector/PerPageSelector';
 
-const App = () => <h1>Pagination</h1>;
+export class App extends React.Component {
+  state = {
+    page: 1,
+    perPage: 5,
+    perPageOptions: [3, 5, 10, 20],
+    withInfo: true,
```
irenhh: what is the purpose of this variable

src/components/Pagination/Pagination.js
```diff
@@ -0,0 +1,91 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import './Pagination.css';
+
+export const Pagination = (props) => {
+  const { total, page, perPage, onPageChange, withInfo } = props;
+
+  const handleButtonClick = pageNumber => onPageChange(pageNumber);
```
irenhh: no need to write this function

src/components/Pagination/Pagination.js
```diff
@@ -0,0 +1,91 @@
...
+
+  const renderButtons = () => (
+    new Array(Math.ceil(total / perPage))
+      .fill('')
+      .map((it, index) => {
+        const buttonClassName = (index + 1 === page)
+          ? 'ui red basic button'
+          : 'ui green basic button';
+
+        const key = index + 1;
```
irenhh: this is almost the same as if you just leave index

src/components/Pagination/Pagination.js
```diff
@@ -0,0 +1,91 @@
...
+            onClick={() => handleButtonClick(index + 1)}
+          >
+            {index + 1}
+          </button>
+        );
+      })
+  );
+
+  const renderInfo = () => {
+    if (!withInfo) {
```
irenhh: when withInfo is false?

src/components/Pagination/Pagination.js
```diff
@@ -0,0 +1,91 @@
...
+
+    const min = page * perPage - perPage + 1;
+    const max = Math.min(page * perPage, total);
+
+    return `${min} - ${max} of ${total}`;
+  };
+
+  const handlePrevButtonClick = () => onPageChange(page - 1);
+
+  const handleNextButtonClick = () => onPageChange(page + 1);
```
irenhh: there is no need to create those methods

src/components/PerPageSelector/PerPageSelector.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import './PerPageSelector.css';
+
+export const PerPageSelector = (props) => {
+  const { perPage, perPageOptions, onPerPageChange } = props;
+
+  const handlePerPageChange = (evt) => {
```
irenhh: event

irenhh: move this function to the place where you call it

src/App.js
AlexanderShpilka: There is a requirement "Add an optional param withInfo to show extra info before the buttons (6 - 10 of 20)". I had no clear idea of what this should look like. I agree that 'withInfo' shouldn't be part of state as there is no way to change it in the app. So I've excluded it from the state and just pass it to the Pagination component as a boolean attribute. If there is no need for additional information one can omit to pass this attribute to the Pagination component.

src/components/Pagination/Pagination.js
AlexanderShpilka: done

src/components/Pagination/Pagination.js
AlexanderShpilka: Now it's false when 'withInfo' isn't passed to the Pagination component.

src/components/Pagination/Pagination.js
AlexanderShpilka: done

src/components/PerPageSelector/PerPageSelector.js
AlexanderShpilka: Discussed that on Slack. Done.

src/components/Pagination/Pagination.js
AlexanderShpilka: Used uuid library for key generation instead of index.

PR #39, opened 2020-05-07
solution

kiramarks:
[**DEMO LINK**](https://kiramarks.github.io/react_pagination/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,6 +1,97 @@
-import React from 'react';
+import React, { Component } from 'react';
+import axios from 'axios';
+import { Posts } from './components/Posts';
+import { Pagination } from './components/Pagination';
+import { SelectSettings } from './components/SelectSettings';
 import './App.css';
 
-const App = () => <h1>Pagination</h1>;
+class App extends Component {
+  state = {
+    posts: [],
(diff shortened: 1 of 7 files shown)


src/components/Posts.js
```diff
@@ -0,0 +1,16 @@
...
+    {posts.map(post => (
+      <li key={post.id} className="list-group-item">
+        {post.title}
+      </li>
+    ))}
+  </ul>
+);
+
+Posts.propTypes = {
+  posts: PropTypes.arrayOf(PropTypes.object).isRequired,
```
irenhh, line 15: describe object every time you use elements from it

src/App.js
```diff
@@ -1,6 +1,97 @@
...
+    if (target.id === 'prev') {
+      currentPage !== 1
+        ? (this.setState(state => ({
+          currentPage: state.currentPage - 1,
+        })))
+        : (this.setState({
+          currentPage: 1,
+        }));
+    } else if (target.id === 'next') {
+      currentPage < +target.parentElement.id
```
irenhh, line 36: when you use parentElement you interact with dom, don't do this in React, there is an option to use arguments for functions

PR #78, opened 2021-05-30
react_pagination

IvanGrekov:
[DEMO LINK](https://IvanGrekov.github.io/react_pagination/)


src/App.js
```diff
@@ -1,6 +1,74 @@
...
+  changePage = pageNumber => (
+    // eslint-disable-next-line consistent-return
+    this.setState((state) => {
+      if (pageNumber !== state.selectedPage) {
+        return {
+          selectedPage: pageNumber,
+        };
+      }
+    })
+  )
```
VinnieJ-2k20, line 21: You are not using previous state directly in calculations, so it is possible to use simple setState here. And try to avoid disabling eslint, this is final measure to use.

```suggestion
  changePage = pageNumber => {
    const { selectedPage } = this.state;
    
    if (pageNumber !== selectedPage) {
      this.setState({
        selectedPage: pageNumber,
      });
    }
  }
```

VinnieJ-2k20 (approved):
Cool job 👍 
Two comments:
1. I added small advice on code optimisation.
2. Try not to use hover effects that alter border-width. It makes the whole structure jump on hover. Better to use color or transforms (like scale).

PR #164, opened 2022-08-05
solution

FANTAZER-NURE:
https://FANTAZER-NURE.github.io/react_pagination/


Bohdan-Maksymiuk (changes requested):
https://user-images.githubusercontent.com/94289038/183690367-c6762e46-8c76-4c2c-a7c4-0cee1a4c5128.mov

fix this bug

PR #223, opened 2022-10-25
pagination was implemented

TheVovchik:
**[DEMO LINK](https://TheVovchik.github.io/react_pagination/)**

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,33 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
+import { Pagination } from './components/Pagination';
 import { getNumbers } from './utils';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
+const total = 42;
+const items = getNumbers(1, total)
   .map(n => `Item ${n}`);
+const options = [3, 5, 10, 20];
 
 export const App: React.FC = () => {
+  const [perPage, setPerPage] = useState(5);
+  const [currentPage, setCurrentPage] = useState(1);
+  const firstItemOnPage = 1 + (currentPage - 1) * perPage;
+  const lastItemOnPage = currentPage * perPage > total
+    ? total
+    : currentPage * perPage;
+
+  const handleOptions = (e: React.ChangeEvent<HTMLSelectElement>) => {
+    setPerPage(+e.target.value);
+    setCurrentPage(1);
+  };
+
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+        {`Page ${currentPage} (items ${firstItemOnPage} - ${lastItemOnPage} of ${total})`}
       </p>
 
       <div className="form-group row">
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -34,62 +50,23 @@ export const App: React.FC = () => {
...
-        <li data-cy="item">Item 1</li>
-        <li data-cy="item">Item 2</li>
-        <li data-cy="item">Item 3</li>
-        <li data-cy="item">Item 4</li>
-        <li data-cy="item">Item 5</li>
+        {items.map((item, index) => {
+          if (index >= (currentPage - 1) * perPage
+            && index < currentPage * perPage) {
+            return (
+              <li key={item} data-cy="item">{item}</li>
```
anastasiia-nahaiska, line 64: Maybe using UUID is a better solution here

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,76 @@
...
+      )}
+      >
+        <a
+          data-cy="prevLink"
+          className="page-link"
+          href="#prev"
+          aria-disabled={currentPage === 1}
+          onClick={() => (currentPage !== 1
+            ? onPageChange(currentPage - 1)
+            : null)}
```
anastasiia-nahaiska, line 34: As I have understood this function will be created every time when will be happened render if we don't create a method
But maybe I`m wrong)

anastasiia-nahaiska (approved):
Cooool!!!

src/components/Pagination/Pagination.tsx
TheVovchik: when we rerender it will dissapear, and on new object we will have new callback
in general it works like your's when you check 
`const getPrevPage = () => {
    if (currentPage !== 1) {
      onPageChange(currentPage - 1);
    }
  };`
  
 maybe in your case you will have undefined

src/App.tsx
```diff
@@ -34,62 +50,23 @@ export const App: React.FC = () => {
...
+      />
       <ul>
-        <li data-cy="item">Item 1</li>
-        <li data-cy="item">Item 2</li>
-        <li data-cy="item">Item 3</li>
-        <li data-cy="item">Item 4</li>
-        <li data-cy="item">Item 5</li>
+        {items.map((item, index) => {
+          if (index >= (currentPage - 1) * perPage
+            && index < currentPage * perPage) {
```
Anna-Tkachenko, line 62: will be better to create separate variables for complicated checks, and use those variables inside the if()

src/App.tsx
Anna-Tkachenko: if you're sure that you'll have unique values in item, you can use it
but 100% sure with idea of using uuid as key for elements

src/components/Pagination/Pagination.tsx
Anna-Tkachenko, line 34: will be better to move this handler in separate function

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,76 @@
...
+            href={`#${page}`}
+            onClick={() => onPageChange(page)}
+          >
+            {page}
+          </a>
+        </li>
+      ))}
+      <li className={classNames(
+        'page-item',
+        { disabled: currentPage === pages.length },
```
Anna-Tkachenko, line 59: you make one check several times, will be better to create separate boolean variable and use it where you need it

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,76 @@
...
+      )}
+      >
+        <a
+          data-cy="nextLink"
+          className="page-link"
+          href="#next"
+          aria-disabled={currentPage === pages.length}
+          onClick={() => (currentPage !== pages.length
+            ? onPageChange(currentPage + 1)
+            : null)}
```
Anna-Tkachenko, line 69: create separate handler for this function

Anna-Tkachenko (approved):

OleksiiBlezniuk (approved):
👍

PR #229, opened 2022-10-28
Implemented the Pagination  as a stateless component to show only the items for a current page

Dikey945:
[DEMO LINK](https://Dikey945.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,43 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
+import { Pagination } from './components/Pagination';
 import { getNumbers } from './utils';
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
-  .map(n => `Item ${n}`);
+const items: number[] = getNumbers(1, 42);
 
 export const App: React.FC = () => {
+  const total = 42;
+  const [perPage, setPerPage] = useState(3);
+  const [currentPage, setCurrentPage] = useState(1);
+  const totalPages = getNumbers(1, Math.ceil(total / perPage));
+  const onChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    event.preventDefault();
+    setPerPage(+event.target.value);
+    setCurrentPage(1);
+  };
+
+  const itemsOnPage = items.filter((item) => (
+    item > currentPage * perPage - perPage
+    && item <= currentPage * perPage
+  ));
+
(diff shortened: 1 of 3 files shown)


iPanda177 (approved):
Cool!)

src/App.tsx
```diff
@@ -1,18 +1,43 @@
...
+import { Pagination } from './components/Pagination';
 import { getNumbers } from './utils';
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
-  .map(n => `Item ${n}`);
+const items: number[] = getNumbers(1, 42);
 
 export const App: React.FC = () => {
+  const total = 42;
```
Anna-Tkachenko, line 9: will be better to move constants out of the component

src/App.tsx
```diff
@@ -1,18 +1,43 @@
...
+  const [perPage, setPerPage] = useState(3);
+  const [currentPage, setCurrentPage] = useState(1);
```
Anna-Tkachenko, line 11: initial default values will be also good to move to the constants

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,97 @@
...
+  currentPage,
+  totalPages,
+  onPageChange,
+}) => {
+  // const isActivePage = (num: number): boolean => num === currentPage;
+  const isPrevPage = currentPage === 1;
+  const isNextPage = currentPage === totalPages.length;
+
+  return (
+    <>
```
Anna-Tkachenko, line 23: react fragment is redundant here

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,97 @@
...
+          </a>
+        </li>
+
+        {totalPages.map((pageNum) => (
+          <li
+            className={classNames(
+              'page-item',
+              { active: pageNum === currentPage },
+            )}
+            key={uuidv4()}
```
Anna-Tkachenko, line 50: you can't call function for generating uuid inside the map callback
in this way you'll have new key on each rendering, it's not help to optimize rendering
you should generate key before rendering list

Anna-Tkachenko (changes requested):

src/components/Pagination/Pagination.tsx
Dikey945: In this case, we have two ul elements with the same nesting, so I can't avoid react fragment usage.

silchyn (approved):
Cool!

PR #234, opened 2022-10-31
done!

dmitriyn3679:
- [DEMO LINK](https://dmitriyn3679.github.io/react_pagination/)

### DIFF
--- a/src/components/Pagination/Pagination.tsx
+++ b/src/components/Pagination/Pagination.tsx
@@ -1,1 +1,88 @@
-export const Pagination = () => {};
+import React from 'react';
+import classNames from 'classnames';
+import { getNumbers } from '../../utils';
+
+type Props = {
+  total: number;
+  perPage: number;
+  currentPage: number;
+  onChangePage: (page: number, maxCountOfPages?: number) => void;
+  items: string[];
+};
+
+export const Pagination: React.FC<Props> = (
+  {
+    total,
+    perPage,
+    currentPage,
+    onChangePage,
+    items,
+  },
+) => {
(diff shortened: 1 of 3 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,88 @@
...
+            data-cy="prevLink"
+            className="page-link"
+            href="#prev"
+            aria-disabled="true"
+          >
+            «
+          </a>
+        </li>
+        {pageItemsList.map(pageNumber => (
+          // eslint-disable-next-line jsx-a11y/click-events-have-key-events,jsx-a11y/no-noninteractive-element-interactions
```
Bohdan-Maksymiuk, line 47: try to google an error and fix it instead of disabling eslint rule. It is a bad practice

Bohdan-Maksymiuk (changes requested):

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,85 @@
...
+          className={classNames(
+            'page-item',
+            { disabled: currentPage === firstPage },
+          )}
+        >
+          <a
+            data-cy="prevLink"
+            className="page-link"
+            href="#prev"
+            aria-disabled="true"
```
TarasHoliuk, line 39: Write here the boolean value (the same as on the 32nd row) - it will fix tests
And the same for `#prev` link

TarasHoliuk (changes requested):
GJ! I've added only one suggestion on how to fix tests

PR #276, opened 2022-12-27
Develop

frontend112:
https://frontend112.github.io/react_pagination/

### DIFF
--- a/src/components/Pagination/Pagination.tsx
+++ b/src/components/Pagination/Pagination.tsx
@@ -1,1 +1,174 @@
-export const Pagination = () => {};
+import { FC } from 'react';
+import cn from 'classnames';
+import './Pagination.scss';
+import { getNumbers } from '../../utils';
+
+interface Props {
+  total: number;
+  perPage: number;
+  currentPage: number;
+  onPageChange: (page: number | string) => void;
+}
+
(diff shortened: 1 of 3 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,115 @@
...
+  perPage,
+  currentPage,
+  onPageChange,
+}) => {
+  const firstItem = (currentPage - 1) * perPage + 1;
+
+  const lastItem = firstItem + perPage > total
+    ? total
+    : firstItem + perPage - 1;
+  const recentItems = getNumbers(firstItem, lastItem).map((n: number) => `Item ${n}`);
```
qoa11a, line 24: You calculate it twice. DRY - don't repeat yourself. Pass it as props and declare it once.

qoa11a (changes requested):
Good job!
One small fix.

src/components/Pagination/Pagination.tsx
frontend112: @qoa11a I thought that I could use only these variables that was descripted in readme that is why I wrote this twice thought that it is not going to pass tests, but it is fine

qoa11a (approved):

PR #279, opened 2022-12-31
react_pagination

ysrckr:
[DEMO LINK](https://ysrckr.github.io/react_pagination/)

### DIFF
--- a/src/components/Pagination/Pagination.tsx
+++ b/src/components/Pagination/Pagination.tsx
@@ -1,1 +1,91 @@
-export const Pagination = () => {};
+import type { MouseEvent, FC } from 'react';
+import cn from 'classnames';
+
+type Props = {
+  total: number;
+  perPage: number;
+  currentPage: number;
+  onPageChange: (action: number | string) => void;
+};
+
+export const Pagination: FC<Props> = ({
+  total,
+  perPage,
+  currentPage,
+  onPageChange,
+}) => {
+  const pages = Math.ceil(total / perPage);
+  const pageNumbers = Array.from({ length: pages }, (_, i) => i + 1);
+
+  const nextClick = (e: MouseEvent) => {
+    e.preventDefault();
+
+    if (currentPage === pages) {
+      return;
+    }
+
+    onPageChange('next');
+  };
+
+  const prevClick = (e: MouseEvent) => {
+    e.preventDefault();
+    if (currentPage === 1) {
+      return;
+    }
+
+    onPageChange('prev');
+  };
+
+  return (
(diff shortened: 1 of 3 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,91 @@
...
+        <a
+          data-cy="prevLink"
+          className="page-link"
+          href="#prev"
+          aria-disabled={currentPage === 1}
+          onClick={e => prevClick(e)}
+        >
+          «
+        </a>
+      </li>
```
meforoma, line 55: let's implement a helper function to generate buttons `« / »`
sort of 
```
const renderStepButton = (direction: 'next' | 'prev') => { ... }
```

meforoma (changes requested):
great job 👏🏼 

let's make buttons to render with a helper function - a step beyond 😉

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,59 @@
...
+  onPageChange: (action: number | string) => void;
+};
+
+export const Pagination: FC<Props> = ({ currentPage, onPageChange, pages }) => {
+  const pageNumbers = Array.from({ length: pages }, (_, i) => i + 1);
+
+  const nextClick = (e: MouseEvent) => {
+    e.preventDefault();
+
+    if (currentPage === pages) {
```
naziamoff, line 17: from its name (`pages`) I expect the object to be there, but it is actually a number

src/components/StepButton/StepButton.tsx
```diff
@@ -0,0 +1,54 @@
+import type { FC, MouseEvent } from 'react';
+import cn from 'classnames';
+
+type Props = {
+  direction: 'next' | 'prev';
```
naziamoff, line 5: better to create enum if we exactly know the string values it's gonna have. That decreases a chance of missing to impossible. Making it a string though, looks kinda hardcoded

src/components/StepButton/StepButton.tsx
```diff
@@ -0,0 +1,54 @@
...
+  handleClick: (e: MouseEvent) => void;
+};
+
+export const StepButton: FC<Props> = ({
+  direction,
+  page,
+  totalPage,
+  handleClick,
+}) => {
+  if (direction === 'prev') {
```
naziamoff, line 17: why not just conditional rendering?
I mean inside just one return

naziamoff (changes requested):
well done 🔥 
Let's polish it a bit though

PR #287, opened 2023-01-06
solution

lefteris-L:
[DEMO LINK](https://lefteris-l.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -47,7 +47,7 @@ export const App: FC = () => {
             className="form-control"
             value={perPage}
             onChange={(event) => {
-              setPerPage(+event.target.value);
+              setPerPage(Number(event.target.value));
               setCurrentPage(1);
             }}
           >
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -8,26 +8,31 @@ const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
 export const App: FC = () => {
-  const [currentPage, setCurrentPage] = useState(1);
-  const [perPage, setPerPage] = useState(5);
+  const defaultItemsPerPage = 5;
+  const defaultPage = 1;
+
+  const [currentPage, setCurrentPage] = useState(defaultPage);
+  const [itemsPerPage, setitemsPerPage] = useState(defaultItemsPerPage);
+
+  const itemsToDisplay = items.slice(0, itemsPerPage);
 
   const [
     visibleItems,
     setVisibleItems,
-  ] = useState(items.slice(0, perPage));
+  ] = useState(itemsToDisplay);
 
   const total = items.length;
-  const from = currentPage * perPage - perPage + 1;
-  const to = Math.min(total, currentPage * perPage);
+  const from = currentPage * itemsPerPage - itemsPerPage + 1;
+  const to = Math.min(total, currentPage * itemsPerPage);
 
   useEffect(() => {
     const newItems = items.slice(
-      (currentPage - 1) * perPage,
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -21,6 +45,11 @@ export const App: React.FC = () => {
             data-cy="perPageSelector"
             id="perPageSelector"
             className="form-control"
+            value={perPage}
+            onChange={(event) => {
+              setPerPage(+event.target.value);
```
meforoma, line 50: use explicit type conversion
```suggestion
              setPerPage(Number(event.target.value));
```

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,94 @@
...
+          aria-disabled={isFirstPage}
+          onClick={() => {
+            if (!isFirstPage) {
+              onPageChange(prev);
+            }
+          }}
+        >
+          «
+        </a>
+      </li>
```
meforoma, line 51: prev and next pages are 95% the same - make a tiny function that will render the proper button
like this:
```
const renderStepButton = (direction: 'next' | 'prev') => {...}
```
and use this short function instead two bulky similar constructs
`{renderStepButton('next')}` / `{renderStepButton('prev')}`

meforoma (changes requested):
please see few suggestions
overall - good job 👏🏼

src/components/Pagination/Pagination.tsx
lefteris-L: That sounds pretty cool. Thank you Roman

src/App.tsx
```diff
@@ -1,18 +1,42 @@
...
+import './App.css';
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
-export const App: React.FC = () => {
+export const App: FC = () => {
+  const [currentPage, setCurrentPage] = useState(1);
+  const [perPage, setPerPage] = useState(5);
```
naziamoff, line 12: per page what? 😄

src/App.tsx
```diff
@@ -1,18 +1,42 @@
...
+
+  const [
+    visibleItems,
+    setVisibleItems,
+  ] = useState(items.slice(0, perPage));
```
naziamoff, line 17: calculating the initial state value inside the useState() looks ugly a bit 🥲 
I'd create a variable for that

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,94 @@
...
+
+  const [prev, setPrev] = useState(currentPage - 1);
+  const [next, setNext] = useState(currentPage + 1);
+
+  useEffect(() => {
+    setNext(currentPage + 1);
+    setPrev(currentPage - 1);
+  }, [currentPage]);
+
+  const renderStepButton = (direction: 'prev' | 'next') => {
```
naziamoff, line 28: - better to separate the type. If you'll need to change it - you will need to look for exactly this piece of code
- if your values type can only be strings - it is much better to use Enum for that. If it is used the way you have it now - the possibility of making a mistake is much higher

src/App.tsx
naziamoff, line 12: 5 looks kinda hardcoded here. It is better to create a variable for that to make it clear

naziamoff (changes requested):
not bad at all. There's a space to improve though, so let's do it)

naziamoff (approved):
wow, great job!

PR #320, opened 2023-02-03
Dev

nzrii-kysel:
[demo](https://nzrii-kysel.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,6 @@
 import { FC, useEffect, useState } from 'react';
 import './App.css';
-import { getNumbers } from './utils';
+import { getExtremes, getNumbers, getVisibleItems } from './utils';
 import { Pagination } from './components/Pagination';
 import { ItemList } from './components/ItemList';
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,96 +1,54 @@
...
+import { PerPageSelector } from './components/PerPageSelector';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+  const page = +(searchParams.get('page') || 1);
+  const perPage = +(searchParams.get('perPage') || 5);
```
denys-cheporniuk, line 16: in the real project, we don't write numbers like this
we create constants for it
```
const CURRENT_PAGE_DEFAULT_VALUE = 1;
const PER_PAGE_DEFAULT_VALUE = 5;
```
and we store these variables in a separate file - constants.ts

denys-cheporniuk (approved):

PR #323, opened 2023-02-06
Pagination was created

anna-inozemets:
[DEMO LINK](https://anna-inozemets.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
+import { Pagination } from './components/Pagination';
 import { getNumbers } from './utils';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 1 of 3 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,87 @@
...
+  perPage,
+  currentPage,
+  onPageChange,
+}) => {
+  const totalPageCount = Math.ceil(total / perPage);
+  const pages: number[] = [];
+
+  for (let i = 1; i <= totalPageCount; i += 1) {
+    pages.push(i);
+  }
```
Avichez, line 21: I think here is better to use function from  `utils` folder to prepare pages.

Avichez (approved):
Look great 🙂 👍

src/App.tsx
```diff
@@ -1,18 +1,31 @@
...
 import './App.css';
+import { Pagination } from './components/Pagination';
 import { getNumbers } from './utils';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const [perPage, setPerPage] = useState(5);
```
alexandra-protyanova, line 11: it is better to save the initial values to constants, to make it easier in future to change them in one place

src/components/Pagination/Pagination.tsx
alexandra-protyanova: agree

alexandra-protyanova (approved):
Well done 👍

PR #324, opened 2023-02-06
solution without react router

Avichez:
This is not the end, planning to implement react router.
[DEMO LINK](https://Avichez.github.io/react_pagination/)


anna-inozemets (approved):
Good and clear solution!)

alexandra-protyanova (approved):
good job

PR #327, opened 2023-02-09
Solution

vlad-boikov:
[DEMO LINK](https://vlad-boikov.github.io/react_pagination/)

### DIFF
--- a/src/components/Pagination/Pagination.tsx
+++ b/src/components/Pagination/Pagination.tsx
@@ -1,1 +1,80 @@
-export const Pagination = () => {};
+import classNames from 'classnames';
+import { getNumbers } from '../../utils';
+
+type Props = {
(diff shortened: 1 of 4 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,80 @@
...
+  const prevPage = currentPage - 1;
+  const nextPage = currentPage + 1;
+
+  const pageChangeHandler = (page: number) => {
+    if (page > pagesQnt || page < 1) {
+      return;
+    }
+
+    onPageChange(page);
+  };
```
DarkMistyRoom, line 33: also you shouldn't call onPageChange if we click on current page

DarkMistyRoom (approved):

PR #329, opened 2023-02-10
Solution

OleksandrOse:
[DEMO LINK](https://OleksandrOse.github.io/react_pagination/)

### DIFF
--- a/src/components/Pagination/Pagination.tsx
+++ b/src/components/Pagination/Pagination.tsx
@@ -32,7 +32,7 @@ export const Pagination: React.FC<Props> = ({
   };
 
   const pageChangeHandler = (page: number) => {
-    if (page > totalPages || page < 1) {
+    if (page > totalPages || page < 1 || page === currentPage) {
       return;
     }
 


Volodymyr-Melnyk-Fron10der (changes requested):
Hi! Great job! But I see you did miss some point here. All jsx on 40-65 lines should be inside <Pagination> component
image

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,102 @@
...
+    return `?page=${nextPage}&perPageSize=${perPage}`;
+  };
+
+  const pageChangeHandler = (page: number) => {
+    if (page > totalPages || page < 1) {
+      return;
+    }
+
+    onPageChange(page);
+  };
```
DarkMistyRoom, line 40: add a query here to avoid calling when we click on the same page

DarkMistyRoom (changes requested):
Good job! Just add one optimization

vlad-demchuk (approved):

PR #345, opened 2023-02-28
Develop

OksanaBaloh:
[DEMO LINK](https://OksanaBaloh.github.io/react_pagination/)

Two commits were added, because solutions slightly different.


TarasHoliuk (approved):
Perfect :+1: 
The both versions are good :fire:

PR #366, opened 2023-03-20
Solution

dissolveyo:
[DEMO LINK](https://dissolveyo.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,17 @@
 import React, { useState } from 'react';
 import './App.css';
+import { PageInfo } from './components/PageInfo';
 import { Pagination } from './components/Pagination';
 import { PerPageSelector } from './components/PerPageSelector';
 import { ItemsCount } from './types/ItemsCount';
-import { getRangeOfItems, createInfo } from './utils/helper';
+import { getRangeOfItems } from './utils/helper';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
+const total = 42;
 
 export const App: React.FC = () => {
   const [currentPage, setCurrentPage] = useState(1);
   const [itemsPerPage, setItemsPerPage] = useState(ItemsCount.Five);
-  const [total] = useState(42);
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,96 +1,48 @@
...
+    = getRangeOfItems(total, currentPage, itemsPerPage);
+
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+        {createInfo(currentPage, firstItem, lastItem, total)}
       </p>
```
OlEzhA0, line 33: as for me, it's should be separate component, with it's own logic

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,110 @@
...
+            aria-disabled={isPageFirst}
+            onClick={handleSetPrevPage}
+          >
+            «
+          </a>
+        </li>
+        {pageItems.map(pageItem => {
+          const isActive = pageItem === currentPage;
+
+          return (
```
OlEzhA0, line 64: why do we need to add new variable if we only use it 1 time?

OlEzhA0 (changes requested):
hey! few minor improvements

src/App.tsx
```diff
@@ -1,96 +1,52 @@
...
+import { PageInfo } from './components/PageInfo';
+import { Pagination } from './components/Pagination';
+import { PerPageSelector } from './components/PerPageSelector';
+import { ItemsCount } from './types/ItemsCount';
+import { getRangeOfItems } from './utils/helper';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
-  .map(n => `Item ${n}`);
+const total = 42;
```
DarkMistyRoom, line 10: would be great to move this total to a separate file with config like you made with ItemsCount. for example `Contstants `file

PR #372, opened 2023-03-28
pagination

Mykola-Busarhin:
[DEMO LINK](https://Mykola-Busarhin.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,34 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const totalPages = 42;
+  const [perPage, setPerPage] = useState(3);
+  const [currentPage, setCurrentPage] = useState(1);
+
+  const firstItem = 1 + (currentPage - 1) * perPage;
+
+  const lastItem = currentPage * perPage > totalPages
+    ? totalPages
+    : currentPage * perPage;
+
+  const handlePerPage = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setPerPage(+event.target.value);
+    setCurrentPage(1);
+  };
+
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+        {`Page ${currentPage} (items ${firstItem} - ${lastItem} of ${totalPages})`}
       </p>
 
       <div className="form-group row">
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -35,62 +54,14 @@ export const App: React.FC = () => {
...
-        <li data-cy="item">Item 5</li>
-      </ul>
+      <Pagination
+        currentPage={currentPage}
+        totalPages={totalPages}
+        perPage={perPage}
+        items={items}
+        onPageChange={setCurrentPage}
+
+      />
```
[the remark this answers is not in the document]
DarkMistyRoom: don't render items inside the Pagination. It is only this section
image

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,98 @@
...
+              key={page}
+              className={classNames('page-item', { active: isChosen })}
+            >
+              <a
+                data-cy="pageLink"
+                className="page-link"
+                href={`#${page}`}
+                onClick={() => {
+                  onPageChange(page);
+                }}
```
DarkMistyRoom, line 62: image

DarkMistyRoom (changes requested):
check [the test](https://github.com/mate-academy/react_pagination/actions/runs/4540780566/jobs/8002125006?pr=372) and fix your code to pass them
image

src/components/Pagination/Pagination.tsx
Mykola-Busarhin: still can`t run my tests in VS Code( If there are still any issues with my code, let me know.

src/App.tsx
```diff
@@ -21,9 +37,10 @@ export const App: React.FC = () => {
             data-cy="perPageSelector"
             id="perPageSelector"
             className="form-control"
+            onChange={handlePerPage}
           >
             <option value="3">3</option>
-            <option value="5">5</option>
+            <option value="5" selected>5</option>
```
etojeDenys, line 43: Instead of using the 'selected' attribute, set the initial value of the select element using the 'value' attribute of the select element. This will prevent the warning in the console. 🌟

src/App.tsx
```diff
@@ -35,61 +52,26 @@ export const App: React.FC = () => {
...
+          const hasfirstItem = index >= (currentPage - 1) * perPage;
+          const haslastItem = index < currentPage * perPage;
+
+          if (hasfirstItem && haslastItem) {
+            return (
+              <li key={item} data-cy="item">{item}</li>
+            );
+          }
+
+          return false;
```
etojeDenys, line 73: Instead of returning 'false', return 'null' for a more appropriate value when not rendering an element in a map function. 😊

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,84 @@
...
+  onPageChange: (page: number) => void;
+};
+
+export const Pagination: React.FC<Props> = ({
+  currentPage,
+  totalPages,
+  perPage,
+  onPageChange,
+}) => {
+  const handleArrowClick = (direction: string, condition: boolean) => {
```
etojeDenys, line 18: Instead of using a string for the direction parameter, consider using a string literal type to restrict the possible values to 'prev' and 'next'.

etojeDenys (changes requested):

vlad-demchuk (approved):

PR #380, opened 2023-04-02
fail

katerinalex:
[DEMO LINK](https://katerinalex.github.io/react_pagination/)


polosanya (changes requested):
Items per page selection doesn't work correctly. Please fix it
pagination

qoa11a (approved):

PR #406, opened 2023-04-29
completed pagination

VenomMB:
[DEMO LINK](https://VenomMB.github.io/react_pagination/)

### DIFF
--- a/src/components/Pagination/Pagination.tsx
+++ b/src/components/Pagination/Pagination.tsx
@@ -1,1 +1,90 @@
-export const Pagination = () => {};
+import classNames from 'classnames';
+import { FC } from 'react';
+
+interface Props {
+  total: number;
+  perPage: number;
+  currentPage: number;
+  onPageChange: (page: number) => void;
+}
+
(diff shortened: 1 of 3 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,90 @@
...
+  const handleClickOnRightButton = () => {
+    if (currentPage < countOfPage) {
+      onPageChange(currentPage + 1);
+    }
+  };
+
+  const handleClickOnLeftButton = () => {
+    if (!isFirstPage) {
+      onPageChange(currentPage - 1);
+    }
```
tiserett, line 36: it works just fine, but if you add more logic inside 'if' block, it would be messy, instead, if current page is first, you can return from function, otherwise execute code below
```suggestion
    if (isFirstPage) {
      return;
    }
    
    onPageChange(currentPage - 1);
```

tiserett (approved):
seems like you was meant to use search params here
anyways, good job
image

PR #411, opened 2023-05-02
Implement pagination

AnyaLyaLya:
[DEMO LINK](https://AnyaLyaLya.github.io/react_pagination/)


tiserett (approved):
url is not working as it should be, sometimes there is wrong data, about page, was also meant, to do this task using searchParams I guess, it's a better approach

PR #412, opened 2023-05-03
Develop

RomeoKs:
[DEMO LINK](https://RomeoKs.github.io/react_pagination/)


src/App.tsx
```diff
@@ -1,18 +1,37 @@
...
 import { getNumbers } from './utils';
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
-  .map(n => `Item ${n}`);
-
 export const App: React.FC = () => {
+  const [itemsPerPage, setItemsPerPage] = useState(5);
+  const [currentPage, setCurrentPage] = useState(1);
+  const [total] = useState(42);
```
meforoma, line 9: no hardcoding pleazzzze ))

and if you're not using stateSetter - why even have a state, right?
regular variable would do.

```suggestion
  const total = currentItems.length;
```

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,80 @@
...
+  const isNextEnabled = !isLastPage;
+
+  return (
+    <ul className="pagination">
+      <li
+        className={classNames('page-item',
+          {
+            disabled: !isPrevEnabled,
+          })}
+      >
```
meforoma, line 33:
```suggestion
  const isPrevDisabled = currentPage === 1;
  const isNextDisabled = isLastPage;

  return (
    <ul className="pagination">
      <li
        className={classNames('page-item',
          {
            disabled: isPrevDisabled,// no need for double negation :)
            // moreover, you could just dump the additional variables
            // since logic is pretty straight
            disabled: currentPage === 1,
            // same with nextPage button
          })}
      >
```

meforoma (approved):
gooddd @RomeoKs 
left a few comments to consider

RomeoKs:
> gooddd @RomeoKs left a few comments to consider

Thank you for suggestions!

PR #413, opened 2023-05-04
Develop

nazar-medushevskyi:
- [DEMO LINK](https://nazar-medushevskyi.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,44 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination/Pagination';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const [itemsToRender, setItemsToRender] = useState(items);
+  const [perPage, setPerPage] = useState(5);
+  const [currentPage, setCurrentPage] = useState(1);
+  const [leadNumbers, setLeadNumbers] = useState('');
+  const total = items.length;
+
+  const onPageChange = (page: number) => {
+    setCurrentPage(() => page);
+  };
+
+  const onPerPageChange = (value: number) => {
+    setPerPage(value);
+    setCurrentPage(() => {
+      return 1;
+    });
+  };
+
+  useEffect(() => {
+    const value1 = (currentPage - 1) * perPage + 1;
+    const value2 = value1 + perPage - 1;
+
+    setItemsToRender(items.slice(value1 - 1, value2));
+    setLeadNumbers(`${value1} - ${value2 > total ? total : value2}`);
+  }, [currentPage, perPage]);
+
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+        {`Page ${currentPage} (items ${leadNumbers} of ${total})`}
       </p>
 
       <div className="form-group row">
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,18 +1,44 @@
...
+  const onPerPageChange = (value: number) => {
+    setPerPage(value);
+    setCurrentPage(() => {
+      return 1;
+    });
+  };
+
+  useEffect(() => {
+    const value1 = (currentPage - 1) * perPage + 1;
+    const value2 = value1 + perPage - 1;
```
polosanya, line 30: It would be great to use more descriptive variables' names

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,68 @@
...
+            () => onPageLinkClick(currentPage - 1, currentPage !== 1)
+          }
+        >
+          «
+        </a>
+      </li>
+
+      {itemsLengthArray.map((_, i) => (
+        <li
+          className={`${(currentPage - 1) === i ? 'active' : ''} page-item`}
```
polosanya, line 39: Don't forget about key attribute

polosanya (changes requested):
Good job! Let's improve your code a bit)

src/App.tsx
```diff
@@ -1,18 +1,44 @@
...
+
+  const handlePageChange = (page: number) => {
+    setCurrentPage(() => page);
+  };
+
+  const handlePerPageChange = (value: number) => {
+    setPerPage(value);
+    setCurrentPage(() => {
+      return 1;
+    });
```
DarkMistyRoom, line 25:
```suggestion
    setCurrentPage(1);
```

src/App.tsx
```diff
@@ -1,18 +1,44 @@
...
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const [itemsToRender, setItemsToRender] = useState(items);
+  const [perPage, setPerPage] = useState(5);
+  const [currentPage, setCurrentPage] = useState(1);
+  const [leadNumbers, setLeadNumbers] = useState('');
```
DarkMistyRoom, line 14: why do you need it in the state? `leadNumbers `can be changed only when other states change

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,70 @@
...
+type Props = {
+  total: number,
+  perPage: number,
+  currentPage: number,
+  onPageChange: (page: number) => void,
+};
+
+export const Pagination: React.FC<Props> = ({
+  total,
+  perPage,
```
DarkMistyRoom, line 12: is it important here? Pagination doesn't matter about perPage, it only needs to know the number of items. So why don't calculate it in the App?

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,70 @@
...
+          }
+        >
+          «
+        </a>
+      </li>
+
+      {itemsLengthArray.map((_, i) => (
+        <li
+          // eslint-disable-next-line react/no-array-index-key
+          key={i}
```
loralevitska, line 40: this rule is written for a reason, you should not disable it, the indexes change. The best option would be to create an object with generated stable identifiers for each number in the prepared array, or at least use the value, for you it is "i + 1"

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,70 @@
...
+            () => onPageLinkClick(currentPage - 1, currentPage !== 1)
+          }
+        >
+          «
+        </a>
+      </li>
+
+      {itemsLengthArray.map((_, i) => (
+        <li
+          // eslint-disable-next-line react/no-array-index-key
```
loralevitska, line 39: Remove comment

src/components/Pagination/Pagination.tsx
loralevitska: not fixed 
and your `itemsLengthArray` can be prepared at App or helper variable

loralevitska (changes requested):
Almost done) Check previous comments

polosanya (changes requested):
There should be numbers instead of `item`. Also, try to fix all failed tests
image
image

OleziO (approved):

PR #422, opened 2023-05-10
solution

LBondarchuk:
[DEMO LINK](https://LBondarchuk.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
+import { Pagination } from './components/Pagination';
 import './App.css';
 import { getNumbers } from './utils';
 
@@ -7,12 +8,53 @@ const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const [selectedValue, setSelectedValue] = useState('5');
+  const [active, setActive] = useState(1);
+  const optionValue = ['3', '5', '10', '20'];
+  const pagesLength = items.length / +selectedValue;
+  const sliceItems = items.slice(
+    +selectedValue * active - +selectedValue, +selectedValue * active,
+  );
+
+  const createPageList = () => {
+    const result = [];
+
+    for (let i = 0; i < pagesLength; i += 1) {
+      result.push(
+        <li
+          key={`#${i + 1}`}
+          className={`page-item ${i + 1 === active ? 'active' : ''}`}
+        >
+          <a
+            data-cy="pageLink"
+            onClick={() => setActive(i + 1)}
+            className="page-link"
+            href={`#${i + 1}`}
+          >
+            {i + 1}
+          </a>
+        </li>,
+      );
+    }
+
+    return result;
+  };
+
+  const itemList = sliceItems.map(item => (
+    <li key={item} data-cy="item">
+      {item}
+    </li>
+  ));
+
+  const pageList = createPageList();
+
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+        {`Page ${active}`}
+        {`(${sliceItems[0].replace('m', 'ms')} - ${sliceItems[sliceItems.length - 1].split(' ')[1]} of ${items.length})`}
       </p>
 
       <div className="form-group row">
@@ -21,11 +63,20 @@ export const App: React.FC = () => {
             data-cy="perPageSelector"
             id="perPageSelector"
             className="form-control"
+            defaultValue={selectedValue}
+            onChange={(e) => {
+              setSelectedValue(e.currentTarget.value);
+              setActive(1);
+            }}
           >
-            <option value="3">3</option>
-            <option value="5">5</option>
-            <option value="10">10</option>
-            <option value="20">20</option>
+            {optionValue.map(value => (
+              <option
+                key={value}
+                defaultValue={value}
+              >
+                {+value}
+              </option>
+            ))}
           </select>
         </div>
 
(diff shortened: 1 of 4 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,51 @@
-export const Pagination = () => {};
+type Props = {
```
polosanya, line 1: The provided props do not match the task description.
image

polosanya (changes requested):
Good job! Let's improve your solution according to the task description

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,74 @@
...
+          </a>
+        </li>,
+      );
+    }
+
+    return result;
+  };
+
+  return (
+    <>
```
tetiana-ivanchenko, line 41: redundant fragment

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,74 @@
...
+      <ul className="pagination">
+        <li
+          className={`page-item ${currentPage === 1 ? 'disabled' : ''}`}
```
tetiana-ivanchenko, line 44: use classnames library

tetiana-ivanchenko (changes requested):

src/App.tsx
```diff
@@ -1,18 +1,26 @@
...
 const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const [selectedValue, setSelectedValue] = useState('5');
+  const [active, setActive] = useState(1);
+  const optionValue = ['3', '5', '10', '20'];
+  const sliceItems = items.slice(
+    +selectedValue * active - +selectedValue, +selectedValue * active,
+  );
```
Anastasiia-Svintsova, line 15: use useMemo()

src/App.tsx
```diff
@@ -21,75 +29,40 @@ export const App: React.FC = () => {
             data-cy="perPageSelector"
             id="perPageSelector"
             className="form-control"
+            defaultValue={selectedValue}
+            onChange={(e) => {
+              setSelectedValue(e.currentTarget.value);
+              setActive(1);
+            }}
```
Anastasiia-Svintsova, line 36: Don't write functions inside JSX. It's better to move it to a separate function

src/App.tsx
```diff
@@ -1,18 +1,26 @@
...
 import { getNumbers } from './utils';
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const [selectedValue, setSelectedValue] = useState('5');
+  const [active, setActive] = useState(1);
+  const optionValue = ['3', '5', '10', '20'];
```
Anastasiia-Svintsova, line 12: Why can't you use just numbers here instead? 
```suggestion
  const [selectedValue, setSelectedValue] = useState(5);
  const [active, setActive] = useState(1);
  const optionValue = [3, 5, 10, 20];
```

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,74 @@
...
+      <li
+        className={cn('page-item', { disabled: currentPage >= pagesLength })}
+      >
+        <a
+          data-cy="nextLink"
+          className="page-link"
+          href="#next"
+          aria-disabled={currentPage >= pagesLength}
+          onClick={() => currentPage < pagesLength
+              && onPageChange(currentPage + 1)}
```
Anastasiia-Svintsova, line 67: It should be disabled in other cases
```suggestion
          onClick={() => onPageChange(currentPage + 1)}
```

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,74 @@
...
+          href="#prev"
+          aria-disabled={currentPage === 1}
+          onClick={() => currentPage !== 1 && onPageChange(currentPage - 1)}
+        >
+          «
+        </a>
+      </li>
+      {createPageList()}
+      <li
+        className={cn('page-item', { disabled: currentPage >= pagesLength })}
```
Anastasiia-Svintsova, line 59: Move this value to a variable, you use it not once `currentPage >= pagesLength`

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,74 @@
...
+  return (
+    <ul className="pagination">
+      <li
+        className={cn('page-item', { disabled: currentPage === 1 })}
+      >
+        <a
+          data-cy="prevLink"
+          className="page-link"
+          href="#prev"
+          aria-disabled={currentPage === 1}
```
Anastasiia-Svintsova, line 51: Same with this value `currentPage === 1`

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,74 @@
...
+
+export const Pagination: React.FC<Props> = ({
+  total,
+  perPage,
+  currentPage,
+  onPageChange,
+}) => {
+  const pagesLength = total / +perPage;
+
+  const createPageList = () => {
```
Anastasiia-Svintsova, line 18: Instead of using this code as a function and adding function calls to JSX, you could refactor it to a constant and directly use it in JSX without the need for function calls.

Anastasiia-Svintsova (changes requested):

src/App.tsx
LBondarchuk: this is the next topic of study, at this stage I do not have knowledge on this topic

src/App.tsx
LBondarchuk: think that it is not necessary here. An additional function will only add more confusion

src/components/Pagination/Pagination.tsx
LBondarchuk: you are very responsible for your work

src/App.tsx
```diff
@@ -1,18 +1,26 @@
...
+  );
+
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+        {`Page ${active} `}
+        {`(${sliceItems[0].replace('m', 'ms').replace('I', 'i')} - ${sliceItems[sliceItems.length - 1].split(' ')[1]} of ${items.length})`}
```
polosanya, line 23: It would be great to create a separate variable for this value as now it looks a bit complicated

polosanya (approved):
Good job!

PR #423, opened 2023-05-11
solution

webdevw:
[DEMO LINK](https://webdevw.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,50 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
-  .map(n => `Item ${n}`);
+const maxItems = 42;
 
 export const App: React.FC = () => {
+  const [currentPage, setCurrentPage] = useState<number>(1);
+  const [perPage, setPerPage] = useState<number>(5);
+  const toMax = perPage * currentPage;
+  const to = toMax < maxItems ? toMax : maxItems;
+  const from = (currentPage - 1) * perPage + 1;
+  const items = getNumbers(from, to).map(n => `Item ${n}`);
+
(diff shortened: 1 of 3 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,106 @@
...
+export const Pagination: React.FC<Props> = ({
+  total,
+  perPage,
+  currentPage,
+  onPageChange,
+}) => {
+  const getNumberOfPages = Math.ceil(total / perPage);
+  const arrayOfPages = getNumbers(1, getNumberOfPages);
+
+  const activePage = (
```
witflash, line 20: Why do you need this function?
You could use `currentPage` instead of calculating `activePage`. 
I don't think if you can have currentPage bigger than the total pages.

witflash (changes requested):
Good job!

But could you check my comment, please?

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,95 @@
...
+  onPageChange: (page:number) => void,
+};
+
+export const Pagination: React.FC<Props> = ({
+  total,
+  perPage,
+  currentPage,
+  onPageChange,
+}) => {
+  const getNumberOfPages = Math.ceil(total / perPage);
```
polosanya, line 17: Consider renaming 'getNumberOfPages' to 'numberOfPages' since it's a variable, not a function.

src/App.tsx
```diff
@@ -1,24 +1,50 @@
...
-        Page 1 (items 1 - 5 of 42)
+        {`Page ${currentPage} (items ${from} - ${to} of ${maxItems})`}
       </p>
 
       <div className="form-group row">
         <div className="col-3 col-sm-2 col-xl-1">
           <select
             data-cy="perPageSelector"
+            value={perPage}
+            onChange={(event) => setItemsPerPage(event)}
```
polosanya, line 47:
```suggestion
            onChange={setItemsPerPage}
```
It will work the same

polosanya (approved):
Good job!

PR #424, opened 2023-05-12
react_pagination

Oleksandr-Maslo:
https://oleksandr-maslo.github.io/react_pagination/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,36 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -21,6 +39,11 @@ export const App: React.FC = () => {
             data-cy="perPageSelector"
             id="perPageSelector"
             className="form-control"
+            value={perPage}
+            onChange={(event) => {
```
VitaliyBondarenko1982, line 43: instead anonymous arrow function create function in component and pass reference here

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,116 @@
...
+            'page-item',
+            { disabled: currentPage === 1 },
+          )
+        }
+        >
+          <a
+            data-cy="prevLink"
+            className="page-link"
+            href="#prev"
+            onClick={() => {
```
VitaliyBondarenko1982, line 34: it is bad practice to pass anonymous arrow functions in jsx.
create handlers and pass reference here and in other places

VitaliyBondarenko1982 (changes requested):

VitaliyBondarenko1982 (approved):

PR #439, opened 2023-05-22
React pagination task was made

SamirMamedov1010:
[DEMO LINK](https://SamirMamedov1010.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useState, useMemo } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
 import { Pagination } from './components/Pagination';
@@ -17,7 +17,6 @@ export const App: React.FC = () => {
   );
 
   const handleChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
-    event.preventDefault();
     setPerPage(+event.target.value);
     setCurrentPage(1);
   };
@@ -25,8 +24,12 @@ export const App: React.FC = () => {
   const lastItem = currentPage * perPage > total
     ? total
     : currentPage * perPage;
+
   const firstItem = (currentPage * perPage) - perPage + 1;
-  const itemsToShow = items.slice(firstItem - 1, lastItem);
+
+  const itemsToShow = useMemo(() => (
+    items.slice(firstItem - 1, lastItem)
+  ), [firstItem, lastItem]);
 
   return (
     <div className="container">
--- a/src/components/Pagination/Pagination.tsx
+++ b/src/components/Pagination/Pagination.tsx
@@ -1,15 +1,14 @@
 import React, { useMemo } from 'react';
 import classNames from 'classnames';
 
-type Props = {
+type PaginationProps = {
   total: number;
   perPage: number;
   currentPage: number;
   onPageChange: (page: number) => void;
-
 };
 
-export const Pagination: React.FC<Props> = ({
+export const Pagination: React.FC<PaginationProps> = ({
   total,
   perPage,
   currentPage,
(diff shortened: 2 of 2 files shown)


src/App.tsx
```diff
@@ -1,18 +1,39 @@
...
+    event.preventDefault();
+    setPerPage(+event.target.value);
+    setCurrentPage(1);
+  };
+
+  const lastItem = currentPage * perPage > total
+    ? total
+    : currentPage * perPage;
+  const firstItem = (currentPage * perPage) - perPage + 1;
+  const itemsToShow = items.slice(firstItem - 1, lastItem);
```
VitaliyBodnar, line 29: you need to use hook useMemo here, because every time when you state was changed method slice() will be called
so it's not okay for performance

src/App.tsx
```diff
@@ -1,18 +1,39 @@
...
+
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+        {`Page ${currentPage} (items ${firstItem} - ${lastItem} of ${total})`}
```
VitaliyBodnar, line 36: you don't need to Template literals here, because React already has data binding for that, like:
Page { currentPage } ( items { firstItem } - { lastItem } of { total }  )

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,119 @@
-export const Pagination = () => {};
+import React, { useMemo } from 'react';
+import classNames from 'classnames';
+
+type Props = {
+  total: number;
+  perPage: number;
+  currentPage: number;
+  onPageChange: (page: number) => void;
+
```
VitaliyBodnar, line 9: delete empty space

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,119 @@
...
+    event.preventDefault();
+
+    if (page <= pageNumbers.length) {
+      onPageChange(page);
+    }
+  };
+
+  const handleNumberClick = (
+    event:
+    React.MouseEvent<HTMLAnchorElement>,
```
VitaliyBodnar, line 50: don't put the interface in the next line

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,119 @@
...
+          'page-item',
+          {
+            disabled: currentPage === 1,
+          },
+        )}
+      >
+        <a
+          data-cy="prevLink"
+          className="page-link"
+          href="#prev"
```
VitaliyBodnar, line 70: let's update it.
In this case, you added "#prev" href for this tag a, and to disable reload page you added event.preventDefault() in handlePrevClick.

You can make it easier, by just using href="javascript:void(0)".
With 'javascript:void(0)' page will not be reloaded, and you don't need to use preventDefault().

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,119 @@
...
+            {
+              active: currentPage === pageNumber,
+            },
+          )}
+          key={pageNumber}
+        >
+          <a
+            data-cy="pageLink"
+            className="page-link"
+            href={`#${pageNumber}`}
```
VitaliyBodnar, line 91: href="javascript:void(0)".

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,119 @@
-export const Pagination = () => {};
+import React, { useMemo } from 'react';
+import classNames from 'classnames';
+
+type Props = {
```
VitaliyBodnar, line 4: try use more describable name for Props type, for example PaginationProps

VitaliyBodnar (changes requested):

src/App.tsx
SamirMamedov1010: Can't make it this way, cause linter say to plays every part of sentence on a new line, and it's less readable(((

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,107 @@
...
+          >
+            {pageNumber}
+          </button>
+        </li>
+      ))}
+      <li className={classNames(
+        'page-item',
+        {
+          disabled: currentPage === pageNumbers.length,
+        },
```
loralevitska, line 92: if the object has only one key-value element, there is no need to add extra lines to move each key to a new line, because it is only one
```suggestion
          { disabled: currentPage === 1 },
        )}
      >
        <button
          type="button"
          data-cy="prevLink"
          className="page-link"
          aria-disabled={currentPage === 1}
          onClick={handlePrevClick}
        >
          «
        </button>
      </li>

      {pageNumbers.map(pageNumber => (
        <li
          className={classNames(
            'page-item',
            { active: currentPage === pageNumber },
          )}
          key={pageNumber}
        >
          <button
            type="button"
            data-cy="pageLink"
            className="page-link"
            onClick={() => handleNumberClick(pageNumber)}
          >
            {pageNumber}
          </button>
        </li>
      ))}
      <li className={classNames(
        'page-item',
        { disabled: currentPage === pageNumbers.length },
```

loralevitska (approved):
good job!

PR #475, opened 2023-06-27
solution

JOWISSA:
[DEMO LINK](https://JOWISSA.github.io/react_pagination/)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,62 @@
...
+            {page}
+          </a>
+        </li>
+      ))}
+      <li className={`page-item ${isNextLinkDisabled ? 'disabled' : ''}`}>
+        <a
+          data-cy="nextLink"
+          className="page-link"
+          href="#next"
+          aria-disabled={nextPage === 0 ? 'true' : 'false'}
```
OleziO, line 54: Replace `nextPage === 0 ? 'true' : 'false'` with `nextPage === 0`. It will be the same

OleziO (changes requested):

sTorba24 (changes requested):
Almost good! 

When i click on `prev` button on last item i returned to 5th page, but it should be 8 for default

demetrxx (approved):
Well done 👍 Go ahead!

PR #476, opened 2023-06-28
added solution

sheva10barca:
[DEMO LINK](https://sheva10barca.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,36 @@
-import React from 'react';
-import './App.css';
-import { getNumbers } from './utils';
+import React, { useState } from 'react';
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
-  .map(n => `Item ${n}`);
+import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
+
+import './App.css';
 
 export const App: React.FC = () => {
+  const [currentPage, setCurrentPage] = useState(1);
+  const [perPage, setPerPage] = useState(5);
+
+  const total = 42;
+
+  const firstItemOnPage = perPage * (currentPage - 1) + 1;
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -21,11 +39,12 @@ export const App: React.FC = () => {
...
             id="perPageSelector"
             className="form-control"
+            onChange={(event) => handleChangePerPage(event)}
           >
-            <option value="3">3</option>
-            <option value="5">5</option>
-            <option value="10">10</option>
-            <option value="20">20</option>
+            <option value={3}>3</option>
+            <option selected value={5}>5</option>
```
VitaliyBondarenko1982, line 45:
```suggestion
            <option value={5}>5</option>
```
instead selected just pass value={perPage} for select

src/App.tsx
```diff
@@ -21,11 +39,12 @@ export const App: React.FC = () => {
...
+            <option value={10}>10</option>
+            <option value={20}>20</option>
```
VitaliyBondarenko1982, line 47:
```suggestion
            {[3, 5, 10, 20].map(option => (
              <option key={option} value={option}>{option}</option>
            )}
```
consider to write in constant array of possoble option and render option with mapping
something like that

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,75 @@
...
+        <a
+          data-cy="nextLink"
+          className="page-link"
+          href="#next"
+          aria-disabled={currentPage === lastPage}
+          onClick={() => {
+            if (currentPage !== lastPage) {
+              onPageChange(currentPage + 1);
+            }
+          }}
```
VitaliyBondarenko1982, line 68: better to make handler in component and pass reference here

VitaliyBondarenko1982 (changes requested):
Good job.
Needs some code improvements.

IvanFesenko (approved):
Well done

PR #484, opened 2023-07-02
Solution

amashutikov:
[DEMO LINK](https://amashutikov.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,44 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
-import { getNumbers } from './utils';
+import { getNumbers, getVisisibleItems } from './utils';
+import { Pagination } from './components/Pagination';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const [itemsPerPage, setItemsPerPage] = useState(5);
+  const [currentPage, setCurrentPage] = useState(1);
+
+  const handlePageChange = (page: number) => {
+    setCurrentPage(page);
+  };
+
+  const handleItemsPerPageChange = (
+    event: React.ChangeEvent<HTMLSelectElement>,
+  ) => {
+    setItemsPerPage(+event.target.value);
+    setCurrentPage(1);
+  };
+
+  const visibleItems = getVisisibleItems(
+    itemsPerPage,
+    currentPage,
+    items.length,
+  );
+
+  const firstVisibleItem = items.indexOf(visibleItems[0]) + 1;
+  const lastVisibleItem = (firstVisibleItem + itemsPerPage - 1) > items.length
+    ? items.length
+    : firstVisibleItem + itemsPerPage - 1;
+
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+        {`Page ${currentPage} (items ${firstVisibleItem} - ${lastVisibleItem} of ${items.length})`}
       </p>
 
       <div className="form-group row">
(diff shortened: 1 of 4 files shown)


src/utils.ts
```diff
@@ -7,3 +7,16 @@ export function getNumbers(from: number, to: number): number[] {
 
   return numbers;
 }
+
+export const getVisisibleItems = (
```
roman-mirzoian, line 11: Here I understand a small typo, be careful with such mistakes. It should be `getVisibleItems` instead of `getVisisibleItems`.

src/utils.ts
```diff
@@ -7,3 +7,16 @@ export function getNumbers(from: number, to: number): number[] {
...
+  itemsPerPage: number,
+  currentPage: number,
+  total: number,
+) => {
+  const lastVisibleItem = currentPage * itemsPerPage;
+
+  return getNumbers(
+    lastVisibleItem - itemsPerPage + 1,
+    (lastVisibleItem > total ? total : lastVisibleItem),
```
roman-mirzoian, line 20: Now everything is written correctly, but the code is a bit difficult to read and maintain. It is better to use the ternary operator outside the arguments of another function and use it as a ready-made result, for example:
```
const indexFrom = lastVisibleItem - itemsPerPage + 1;
const indexTo= lastVisibleItem > total ? total : lastVisibleItem;
getNumbers(indexFrom, indexTo);
```

src/App.tsx
```diff
@@ -34,62 +62,21 @@ export const App: React.FC = () => {
...
       <ul>
-        <li data-cy="item">Item 1</li>
-        <li data-cy="item">Item 2</li>
-        <li data-cy="item">Item 3</li>
-        <li data-cy="item">Item 4</li>
-        <li data-cy="item">Item 5</li>
+        {visibleItems.map(item => (
+          <li
+            data-cy="item"
+          >
```
roman-mirzoian, line 76: Do not forget that when we work with arrays in this way, we have to set a `key` property for each element.

src/utils.ts
```diff
@@ -7,3 +7,16 @@ export function getNumbers(from: number, to: number): number[] {
```
roman-mirzoian, line 7: Now everything is written correctly, but try not to shorten the naming to unreadable names and use more understandable names ( `n` -> `num` / `number` ).

roman-mirzoian (changes requested):
The job is well done and just needs a bit of work on the readability of the code.

src/App.tsx
```diff
@@ -1,18 +1,44 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
-import { getNumbers } from './utils';
+import { getNumbers, getVisibleItems } from './utils';
+import { Pagination } from './components/Pagination';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
naziamoff, line 6: do you still need this linting rule disabling?

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,80 @@
...
+
+export const Pagination: React.FC<Props> = ({
+  total,
+  perPage,
+  currentPage,
+  onPageChange,
+}) => {
+  const numberOfPageItems = Math.ceil(total / perPage);
+
+  const paginations = getNumbers(1, numberOfPageItems);
```
naziamoff, line 19: what's paginations?

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,80 @@
...
+  onPageChange: (page: number) => void;
+};
+
+export const Pagination: React.FC<Props> = ({
+  total,
+  perPage,
+  currentPage,
+  onPageChange,
+}) => {
+  const numberOfPageItems = Math.ceil(total / perPage);
```
naziamoff, line 17: can't get what's stored inside this variable, better to come up with something more clear for the reader

naziamoff (approved):
looks good. However, you should be more attentive about the naming, as not all of your names are clear and easy to read

PR #488, opened 2023-07-03
Basic solution

adv555:
**[DEMO LINK](https://adv555.github.io/react_pagination/)**

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
 import { useSearchParams } from 'react-router-dom';
 import './App.css';
 import { getNumbers } from './utils';
@@ -8,11 +8,19 @@ const items = getNumbers(1, 42);
 
 export const App: React.FC = () => {
   const total = items.length;
+  const countOptions = [3, 5, 10, 20];
 
   const [searchParams, setSearchParams] = useSearchParams();
-
   const [perPage, setPerPage] = useState(5);
   const [currentPage, setCurrentPage] = useState(1);
+
+  useEffect(() => {
+    searchParams.set('page', '2');
+    searchParams.set('perPage', '7');
+
+    setSearchParams(searchParams);
+  },
+  []);
 
   const fromItem = (currentPage - 1) * perPage + 1;
   const toItem = currentPage * perPage > total ? total : currentPage * perPage;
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,18 +1,34 @@
...
 import { getNumbers } from './utils';
-
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
-  .map(n => `Item ${n}`);
+import { Pagination } from './components/Pagination';
 
 export const App: React.FC = () => {
+  const total = 42;
+
```
MaxSchmide, line 8:
```suggestion
  const total = items.length;

```

MaxSchmide: use dynamic variable instead of static

MaxSchmide (approved):
Looks good 👍

src/App.tsx
```diff
@@ -1,18 +1,49 @@
...
+  };
+
+  const handlePageChange = (page: number) => {
+    setCurrentPage(page);
+
+    searchParams.set('page', page.toString());
+    searchParams.set('perPage', perPage.toString());
+
+    setSearchParams(searchParams);
+  };
```
YegorKorenkov, line 39: You set query params but not apply them on page load. After reloading this params should set to state
image

src/App.tsx
```diff
@@ -21,11 +52,14 @@ export const App: React.FC = () => {
...
             id="perPageSelector"
             className="form-control"
+            onChange={handleSelect}
+            defaultValue={perPage}
           >
-            <option value="3">3</option>
-            <option value="5">5</option>
-            <option value="10">10</option>
-            <option value="20">20</option>
+            {[3, 5, 10, 20].map(option => (
```
YegorKorenkov, line 58: Better to move to separate variable
```suggestion
            {countOptions.map(option => (
```

YegorKorenkov (changes requested):

src/App.tsx
adv555: fixed ✅

src/App.tsx
adv555: I'm not sure that I understood correctly what it should be, but set `page=2&perPage=7` in the URL and apply them on page load. 
A bit confusing why 2 and 7🤔 if first page=1 perPage=5

YegorKorenkov: `page=2&perPage=7` it's a example. It should work like this - when the page loads you take the query parameters from the url and write them into the appropriate state variables

YegorKorenkov (approved):

PR #490, opened 2023-07-04
solution

IzzetYusufov:
[DEMO LINK] (https://IzzetYusufov.github.io/react_pagination/)

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -1,1 +1,3 @@
-/* not empty */
+/* iframe {
+  display: none;
+} */
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,96 +1,50 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
+import { Select } from './components/Select';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
-  .map(n => `Item ${n}`);
+const items = getNumbers(1, 42).map(n => `Item ${n}`);
+
+function itemsPerPage(perPageNum: number) {
+  const newItems = [...items];
+  const resultItems = [];
+
+  for (let i = 0; i < newItems.length; i += perPageNum) {
+    resultItems.push(newItems.slice(i, (i + perPageNum)));
+  }
+
+  return resultItems;
+}
 
 export const App: React.FC = () => {
+  const [page, setPage] = useState<number>(1);
+  const [perPage, setPerPage] = useState(5);
+
+  const itemsToDisplay = itemsPerPage(perPage);
+  const itemsFrom = [...itemsToDisplay[page - 1]].shift() as string;
+  const itemsTo = [...itemsToDisplay[page - 1]].pop() as string;
+
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+        {`Page ${page} (items ${+itemsFrom.slice(-2)} - ${+itemsTo.slice(-2)} of ${items.length})`}
       </p>
 
-      <div className="form-group row">
-        <div className="col-3 col-sm-2 col-xl-1">
-          <select
-            data-cy="perPageSelector"
-            id="perPageSelector"
-            className="form-control"
-          >
-            <option value="3">3</option>
-            <option value="5">5</option>
-            <option value="10">10</option>
-            <option value="20">20</option>
-          </select>
-        </div>
+      <Select
+        selectValue={perPage}
+        setOptionPage={setPerPage}
+        ressetPage={setPage}
+      />
 
-        <label htmlFor="perPageSelector" className="col-form-label col">
-          items per page
-        </label>
-      </div>
-
-      {/* Move this markup to Pagination */}
-      <ul className="pagination">
-        <li className="page-item disabled">
-          <a
-            data-cy="prevLink"
-            className="page-link"
-            href="#prev"
-            aria-disabled="true"
-          >
-            «
-          </a>
-        </li>
-        <li className="page-item active">
-          <a data-cy="pageLink" className="page-link" href="#1">1</a>
-        </li>
-        <li className="page-item">
-          <a data-cy="pageLink" className="page-link" href="#2">2</a>
-        </li>
-        <li className="page-item">
-          <a data-cy="pageLink" className="page-link" href="#3">3</a>
-        </li>
-        <li className="page-item">
-          <a data-cy="pageLink" className="page-link" href="#4">4</a>
-        </li>
-        <li className="page-item">
-          <a data-cy="pageLink" className="page-link" href="#5">5</a>
-        </li>
-        <li className="page-item">
-          <a data-cy="pageLink" className="page-link" href="#6">6</a>
-        </li>
-        <li className="page-item">
-          <a data-cy="pageLink" className="page-link" href="#7">7</a>
-        </li>
-        <li className="page-item">
-          <a data-cy="pageLink" className="page-link" href="#8">8</a>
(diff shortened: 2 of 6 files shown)


src/components/Select/Select.tsx
```diff
@@ -0,0 +1,41 @@
+type Props = {
+  selectValue: number,
+  setOptionPage: (value: number) => void,
+  ressetPage: (page: number) => void,
```
volodymyr-soltys97, line 4: There's a typo in the property name 'ressetPage'. It should be 'resetPage'.

src/components/Select/Select.tsx
```diff
@@ -0,0 +1,41 @@
...
+  <div className="form-group row">
+    <div className="col-3 col-sm-2 col-xl-1">
+      <select
+        data-cy="perPageSelector"
+        id="perPageSelector"
+        className="form-control"
+        value={selectValue}
+        onChange={(event) => {
+          setOptionPage(+event.target.value);
+          ressetPage(1);
```
volodymyr-soltys97, line 23: There's a typo in the function name 'ressetPage'. It should be 'resetPage'.

src/components/Select/Select.tsx
```diff
@@ -0,0 +1,41 @@
...
+  setOptionPage: (value: number) => void,
+  ressetPage: (page: number) => void,
+};
+
+const selectOption = [3, 5, 10, 20];
+
+export const Select: React.FC<Props> = ({
+  selectValue,
+  setOptionPage,
+  ressetPage,
```
volodymyr-soltys97, line 12: There's a typo in the function name 'ressetPage'. It should be 'resetPage'.

src/components/Select/Select.tsx
```diff
@@ -0,0 +1,41 @@
+type Props = {
+  selectValue: number,
+  setOptionPage: (value: number) => void,
+  ressetPage: (page: number) => void,
+};
+
+const selectOption = [3, 5, 10, 20];
```
volodymyr-soltys97, line 7: The array 'selectOption' could be better named as 'selectOptions' for clarity.

src/App.tsx
```diff
@@ -1,96 +1,50 @@
...
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
-  .map(n => `Item ${n}`);
+const items = getNumbers(1, 42).map(n => `Item ${n}`);
+
+function itemsPerPage(perPageNum: number) {
+  const newItems = [...items];
+  const resultItems = [];
+
+  for (let i = 0; i < newItems.length; i += perPageNum) {
```
volodymyr-soltys97, line 14: You are using a 'for' loop to create a new array. It's better to use 'map' method in this case.

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,102 @@
...
+            href="#next"
+            aria-disabled={rightArrowDisable}
+            onClick={handleCellClick}
+          >
+            »
+          </a>
+        </li>
+      </ul>
+      <ul>
+        {itemsToDisplay[currentPage - 1].map(item => (
```
volodymyr-soltys97, line 91: This line will throw an error if itemsToDisplay[currentPage - 1] is undefined. Add a check to ensure that itemsToDisplay[currentPage - 1] is defined before calling map on it.

src/App.css
```diff
@@ -1 +1,3 @@
-/* not empty */
+/* iframe {
+  display: none;
+} */
```
volodymyr-soltys97, line 3: Remove comments

volodymyr-soltys97 (changes requested):
Good job 👍 

src/App.css
```diff
@@ -1 +0,0 @@
-/* not empty */
```
anastasiiavorobiova, line 1: The CSS file is now empty. Make sure to add necessary styles for your `Pagination` component.

src/components/Select/Select.tsx
```diff
@@ -0,0 +1,41 @@
+type Props = {
+  selectValue: number,
+  setOptionPage: (value: number) => void,
+  resetPage: (page: number) => void,
+};
+
+const selectOptions = [3, 5, 10, 20];
```
anastasiiavorobiova, line 7: The array of select options should be a prop to make the component more reusable. If it's defined inside the component, you will be unable to change the options without modifying the component itself.

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,104 @@
...
+  itemsToDisplay: Array<string>[],
+  onPageChange: (page: number) => void,
+};
+
+export const Pagination = ({
+  currentPage,
+  itemsToDisplay,
+  onPageChange,
+}: Props) => {
+  const qntyOfCells = getNumbers(1, itemsToDisplay.length);
```
anastasiiavorobiova, line 16: It would be more performant to calculate the quantity of pages once when the component mounts or when the total or perPage props change, instead of on every render.

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,104 @@
...
+      return onPageChange(currentPage - 1);
+    }
+
+    if (leftRightArrow === '#next') {
+      return onPageChange(currentPage + 1);
+    }
+
+    const cellValue = +event.currentTarget.innerText;
+
+    return cellValue !== currentPage && onPageChange(cellValue);
```
anastasiiavorobiova, line 33: The condition `cellValue !== currentPage` can be confusing. It's recommended to wrap it in parentheses for better readability.

src/App.tsx
```diff
@@ -1,96 +1,52 @@
...
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
-  .map(n => `Item ${n}`);
+const items = getNumbers(1, 42).map(n => `Item ${n}`);
+
+function itemsPerPage(perPageNum: number) {
+  const newItems = [...items];
+  const resultItems: Array<string>[] = [];
+
+  newItems.map((_item, i) => {
```
anastasiiavorobiova, line 14: The 'map' function is not used correctly here. It is used to transform every element of the original array into something else, returning a new array. Here, you are using it like 'forEach' or 'reduce'. You should replace it if you don't want to transform the array elements.

src/App.tsx
```diff
@@ -1,96 +1,52 @@
...
+      ? resultItems.push([...items].slice(i, (i + perPageNum)))
+      : null;
+  });
+
+  return resultItems;
+}
 
 export const App: React.FC = () => {
+  const [page, setPage] = useState<number>(1);
+  const [perPage, setPerPage] = useState(5);
```
anastasiiavorobiova, line 25: Consider adding a generic type here

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,104 @@
...
+  onPageChange,
+}: Props) => {
+  const qntyOfCells = getNumbers(1, itemsToDisplay.length);
+  const leftArrowDisable = currentPage === 1;
+  const rightArrowDisable = currentPage === itemsToDisplay.length;
+
+  const handleCellClick = (event: React.MouseEvent<HTMLAnchorElement>) => {
+    const leftRightArrow = event.currentTarget.getAttribute('href');
+
+    if (leftRightArrow === '#prev') {
```
anastasiiavorobiova, line 23: Avoid using magic strings. Consider using constants with meaningful names

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,104 @@
...
+      return onPageChange(currentPage - 1);
+    }
+
+    if (leftRightArrow === '#next') {
```
anastasiiavorobiova, line 27: Avoid using magic strings. Consider using constants with meaningful names

volodymyr-soltys97 (approved):
Good job 👍

PR #493, opened 2023-07-04
pagination

Ir-ra:
[DEMO LINK](https://Ir-ra.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,39 @@
-import React from 'react';
+import React, { ChangeEvent, useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const [perPage, setPerPage] = useState(5);
+  const [currentPage, setCurrentPage] = useState(1);
+
+  const handlePerPage = (event: ChangeEvent<HTMLSelectElement>) => {
+    setPerPage(+event.target.value);
+    setCurrentPage(1);
+  };
+
+  const handleCurrentPage = (page: number) => {
+    if (currentPage !== page) {
+      setCurrentPage(page);
+    }
+  };
+
+  const startItem = ((currentPage - 1) * perPage);
+  const endItem = (currentPage * perPage) > items.length
+    ? items.length
+    : startItem + perPage;
+  const arrOfPerPageItems = items.slice(startItem, endItem);
+
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+        {`Page ${currentPage} (items ${startItem + 1} - ${endItem} of ${items.length})`}
       </p>
 
       <div className="form-group row">
@@ -21,6 +42,8 @@ export const App: React.FC = () => {
             data-cy="perPageSelector"
             id="perPageSelector"
             className="form-control"
+            value={perPage}
+            onChange={(event) => handlePerPage(event)}
           >
             <option value="3">3</option>
             <option value="5">5</option>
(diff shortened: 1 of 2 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,89 @@
...
+  onPageChange,
+}) => {
+  const handleSelectPage = (page: number) => {
+    if (currentPage !== page) {
+      onPageChange(page);
+    }
+  };
+
+  const handleClickRight = () => {
+    if (currentPage !== 0) {
```
polosanya, line 26: The condition in the 'handleClickRight' function is incorrect. You should check if the current page isn't the last one, not if it's not zero.

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,89 @@
...
+
+        {arrOfPages.map((page, i) => {
+          return (
+            <li className={classNames('page-item',
+              { active: currentPage === i + 1 })}
+            >
+              <a
+                data-cy="pageLink"
+                className="page-link"
+                key={page.slice(5)}
```
polosanya, line 65: The 'key' prop should be on the top-level element inside map(), not on the child elements. Move the 'key' prop to the 'li' element.

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,89 @@
-export const Pagination = () => {};
+import React from 'react';
+import classNames from 'classnames';
+
+type Props = {
+  items: string[];
```
polosanya, line 5: Items prop is missing in the task description
image

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,89 @@
...
+
+  const handleClickRight = () => {
+    if (currentPage !== 0) {
+      onPageChange(currentPage + 1);
+    }
+  };
+
+  const handleClickLeft = () => {
+    if (currentPage !== Math.ceil(total / perPage) + 1) {
+      onPageChange(currentPage - 1);
```
polosanya, line 33: Same here

polosanya (changes requested):
Let's improve your code a bit

src/components/Pagination/Pagination.tsx
Ir-ra: Don't understand, if I need items in Pagination component, how I could take them in different way?

loralevitska: Pagination should separate from list. List of items move to another component

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,92 @@
...
+    }
+  };
+
+  const leftDisabled = currentPage === 1;
+  const rightDisabled = currentPage === lastPage;
+
+  const arrOfPages = items.slice(0, lastPage);
+
+  return (
+    <>
```
loralevitska, line 45: Remove redundant [`react.fragment`](https://www.memberstack.com/blog/react-fragment)

loralevitska (changes requested):
Almost done) Let's improve you code with last unresolve comment. Scroll it and check hints or if you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review
image

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,88 @@
...
+  onPageChange,
+}) => {
+  const handleSelectPage = (page: number) => {
+    if (currentPage !== page) {
+      onPageChange(page);
+    }
+  };
+
+  const lastPage = Math.ceil(total / perPage);
+  const arrOfPages = getNumbers(1, lastPage).map(n => n);
```
Anastasiia-Svintsova, line 25: The map function is unnecessary here. getNumbers() function already returns an array so you can directly assign it to arrOfPages like this: const arrOfPages = getNumbers(1, lastPage);

src/App.tsx
```diff
@@ -1,18 +1,39 @@
-import React from 'react';
+import React, { ChangeEvent, useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
Anastasiia-Svintsova, line 6: You should remove all eslint disabling. It's better to fix errors

Anastasiia-Svintsova (approved):

PR #500, opened 2023-07-05
React__pagination

Soi4An:
DEMO - https://Soi4An.github.io/react_pagination/

Test's problems:
Prev link:
- should be enabled if the 1st page is not selected

Next link:
- should be disabled when the last page is selected
- should not change the page if it is already the last

but these is working corret, may you know where problems?

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,51 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination/Pagination';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
+export enum Steps {
+  Three = 3,
+  Five = 5,
+  Ten = 10,
+  Twenty = 20,
+}
+
+const allSteps: Steps[] = [3, 5, 10, 20];
+const totalNumbersPoints = items.length;
+const DEF_PAGE_STEP: Steps = 5;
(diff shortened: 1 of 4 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,71 @@
...
+    <ul className="pagination">
+      <li className={currentPage === 1
+        ? 'page-item disabled'
+        : 'page-item'}
+      >
+        <a
+          data-cy="prevLink"
+          className="page-link"
+          href="#prev"
+          aria-disabled="true"
```
mykhalenych, line 30: change false into if first page is active

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,71 @@
...
+
+      <li className={currentPage === quantityPages
+        ? 'page-item disabled'
+        : 'page-item'}
+      >
+        <a
+          data-cy="nextLink"
+          className="page-link"
+          href="#next"
+          aria-disabled="false"
```
mykhalenych, line 63: change false into if last page is active

src/App.tsx
```diff
@@ -21,11 +54,16 @@ export const App: React.FC = () => {
             data-cy="perPageSelector"
             id="perPageSelector"
             className="form-control"
+            defaultValue={DEF_PAGE_STEP}
+            onChange={(event) => {
```
mykhalenych, line 58: create separate function

mykhalenych (changes requested):

nazarmatsevych (approved):

PR #501, opened 2023-07-05
task done

pos1bl:
https://pos1bl.github.io/react_pagination/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,28 @@
-import React from 'react';
+import React, { useState } from 'react';
+
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const total = 42;
+  const [perPage, setPerPage] = useState(5);
+  const [curPage, setCurPage] = useState(1);
+  const startValue = perPage * curPage - (perPage - 1);
+  const endValue = perPage * curPage <= total
+    ? perPage * curPage
+    : total;
+
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+        {`Page ${curPage} (items ${startValue} - ${endValue} of ${total})`}
       </p>
 
       <div className="form-group row">
@@ -21,9 +31,13 @@ export const App: React.FC = () => {
             data-cy="perPageSelector"
             id="perPageSelector"
             className="form-control"
+            onChange={(event) => {
+              setPerPage(+event.target.value);
+              setCurPage(1);
+            }}
           >
             <option value="3">3</option>
-            <option value="5">5</option>
+            <option value="5" selected>5</option>
             <option value="10">10</option>
             <option value="20">20</option>
           </select>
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -21,9 +31,13 @@ export const App: React.FC = () => {
...
             id="perPageSelector"
             className="form-control"
+            onChange={(event) => {
+              setPerPage(+event.target.value);
+              setCurPage(1);
+            }}
           >
             <option value="3">3</option>
-            <option value="5">5</option>
+            <option value="5" selected>5</option>
```
agniya-i, line 40: Not necessary to use the `selected` attribute directly on an `<option>`, since the `value` attribute is specified on the parent `<select>`.

src/App.tsx
```diff
@@ -34,62 +48,25 @@ export const App: React.FC = () => {
...
+        }}
+      />
+
       <ul>
-        <li data-cy="item">Item 1</li>
-        <li data-cy="item">Item 2</li>
-        <li data-cy="item">Item 3</li>
-        <li data-cy="item">Item 4</li>
-        <li data-cy="item">Item 5</li>
+        {[...items.slice(startValue - 1, endValue)]
```
agniya-i, line 61: Avoid using the spread operator to clone arrays. It can lead to performance issues for large arrays. Consider using the 'slice' method directly instead.

Also, it would be nice to create separate for this value

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,88 @@
...
+
+export const Pagination: React.FC<Props> = ({
+  total,
+  perPage,
+  currentPage,
+  onPageChange,
+}) => {
+  const pagesNumber = Math.ceil(total / perPage);
+  const pages = getNumbers(1, pagesNumber);
+  const ifFirst = () => currentPage === 1;
```
agniya-i, line 20: Instead of creating new functions inside the component for checking currentPage, you can just do the comparisons inline - it's enough in this case

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,88 @@
...
+          className="page-link"
+          href="#prev"
+          aria-disabled={ifFirst()}
+        >
+          «
+        </a>
+      </li>
+
+      {pages.map(page => (
+        <li className={classNames(
```
agniya-i, line 47: Each child in a list should have a unique 'key' prop.

src/App.tsx
```diff
@@ -1,18 +1,28 @@
...
+import { Pagination } from './components/Pagination';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const total = 42;
+  const [perPage, setPerPage] = useState(5);
+  const [curPage, setCurPage] = useState(1);
```
agniya-i, line 14:
```suggestion
  const [currentPage, setCurrentPage] = useState(1);
```
Don't forget to use more descriptive variables names

agniya-i (changes requested):

src/components/Pagination/Pagination.tsx
pos1bl: I created a new function, because i need to check currentPage 3 times. To better understanding and if in future i wll need to change the logic of my code, its better to use function

src/App.tsx
pos1bl: I dont understand this comment. If i reload page, the selceted option shoud be "5" as in the sample

src/App.tsx
```diff
@@ -1,18 +1,29 @@
...
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const total = 42;
```
DarkMistyRoom, line 12: move it out of the component. It's redundant here

src/App.tsx
```diff
@@ -34,62 +49,24 @@ export const App: React.FC = () => {
...
-        <li data-cy="item">Item 4</li>
-        <li data-cy="item">Item 5</li>
+        {pagesPerPage.map(item => (
+          <li
+            data-cy="item"
+            key={item}
+          >
+            {item}
+          </li>
+        ))}
```
DarkMistyRoom, line 69: would be better to create `List` component and move it there

src/App.tsx
```diff
@@ -34,62 +49,24 @@ export const App: React.FC = () => {
...
-          </a>
-        </li>
-      </ul>
+      <Pagination
+        total={total}
+        perPage={perPage}
+        currentPage={currentPage}
+        onPageChange={(page: number) => {
+          setCurrentPage(page);
+        }}
```
DarkMistyRoom, line 58:
```suggestion
        onPageChange={setCurrentPage}
```

DarkMistyRoom: it's a short version

src/components/Pagination/Pagination.tsx
DarkMistyRoom: it is ok for `ifCur` because it's taken an argument. But why don't do just
`const ifFirst = currentPage === 1;`?

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,90 @@
...
+export const Pagination: React.FC<Props> = ({
+  total,
+  perPage,
+  currentPage,
+  onPageChange,
+}) => {
+  const pagesNumber = Math.ceil(total / perPage);
+  const pages = getNumbers(1, pagesNumber);
+  const ifFirst = () => currentPage === 1;
+  const ifCur = (thisPage: number) => currentPage === thisPage;
```
DarkMistyRoom, line 21:
```suggestion
  const isCurrent = (thisPage: number) => currentPage === thisPage;
```

DarkMistyRoom (changes requested):

src/App.tsx
pos1bl: image

src/App.tsx
pos1bl: image

mykhalenych (approved):

PR #504, opened 2023-07-06
react_pagination

uliana-kudriavtseva:
- [DEMO LINK](https://uliana-kudriavtseva.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,36 @@
-import React from 'react';
-import './App.css';
+import { useState } from 'react';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
+import './App.css';
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
+const selectValues = [3, 5, 10, 20];
+
 export const App: React.FC = () => {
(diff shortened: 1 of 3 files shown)


sTorba24 (changes requested):
Almost good! 

When i change number of items per page on paging > 1 i don't see items
image

src/App.tsx
```diff
@@ -1,18 +1,36 @@
...
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
+import './App.css';
+
+const items = getNumbers(1, 42);
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
-  .map(n => `Item ${n}`);
+const selectValues = [3, 5, 10, 20];
```
OleziO, line 8: Better is to use object for this case.

src/App.tsx
```diff
@@ -1,18 +1,36 @@
...
 
 export const App: React.FC = () => {
+  const [currentPage, setCurrentPage] = useState(1);
```
OleziO, line 11: Create object with page num and itemsPerPage values by default.

OleziO (changes requested):

VitaliyBondarenko1982 (approved):
Well done

PR #529, opened 2023-07-13
solution

myplekan:
[DEMO LINK](https://myplekan.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,35 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
+const maxItems = 42;
+const items = getNumbers(1, maxItems)
   .map(n => `Item ${n}`);
+const options = [3, 5, 10, 20];
 
 export const App: React.FC = () => {
+  const [currentPage, setCurrentPage] = useState(1);
+  const [perPage, setPerPage] = useState(5);
+
+  const firstItem = currentPage * perPage - perPage;
+  const lastItem = currentPage * perPage > maxItems
+    ? maxItems
+    : currentPage * perPage;
+
+  const pageItems = items.slice(firstItem, lastItem);
+
+  const handlePerPageChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setPerPage(+event.target.value);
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,18 +1,35 @@
...
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
+const maxItems = 42;
+const items = getNumbers(1, maxItems)
   .map(n => `Item ${n}`);
+const options = [3, 5, 10, 20];
```
OleziO, line 9: Use object instead of arr in this case.

OleziO (changes requested):

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,86 @@
...
+
+export const Pagination: React.FC<Props> = ({
+  total,
+  perPage,
+  currentPage,
+  onPageChange,
+}) => {
+  const numberPag = Math.ceil(total / perPage);
+
+  function countPagination() {
```
volodymyr-soltys97, line 19: It's more efficient to use useMemo for the 'countPagination' function to avoid unnecessary re-rendering.

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,86 @@
...
+    for (let i = 1; i <= numberPag; i += 1) {
+      countArr.push(i);
+    }
+
+    return countArr;
+  }
+
+  const count = countPagination();
+
+  function changeCheck(numPage: number) {
```
volodymyr-soltys97, line 31: The 'changeCheck' function can be replaced with a useCallback hook to avoid unnecessary re-rendering.

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,86 @@
...
+  onPageChange: (page: number) => void,
+}
+
+export const Pagination: React.FC<Props> = ({
+  total,
+  perPage,
+  currentPage,
+  onPageChange,
+}) => {
+  const numberPag = Math.ceil(total / perPage);
```
volodymyr-soltys97, line 17:
```suggestion
  const numberPage = Math.ceil(total / perPage);
```

volodymyr-soltys97 (changes requested):
Good job 👍 

volodymyr-soltys97 (approved):
Nice work

PR #546, opened 2023-07-21
add solution

denis-kononenko:
[DEMO LINK](https://denis-kononenko.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,40 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const [currentPage, setCurrentPage] = useState<number>(1);
+  const [perPage, setPerPage] = useState<number>(5);
+
+  const startIndex = (currentPage - 1) * perPage;
+  let endIndex = currentPage * perPage;
+
+  if (endIndex > items.length) {
+    endIndex = items.length;
+  }
+
+  const visibleItems = items.slice(startIndex, endIndex);
+
+  const selectPerPage = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    const selectedValue = Number(event.target.value);
+
+    if (selectedValue !== perPage) {
+      setPerPage(selectedValue);
+      setCurrentPage(1);
+    }
+  };
+
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+        {`Page ${currentPage} (items ${startIndex + 1} - ${endIndex} of ${items.length})`}
       </p>
 
       <div className="form-group row">
@@ -21,6 +43,8 @@ export const App: React.FC = () => {
             data-cy="perPageSelector"
             id="perPageSelector"
             className="form-control"
+            onChange={(e) => selectPerPage(e)}
+            value={perPage}
           >
             <option value="3">3</option>
             <option value="5">5</option>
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,18 +1,39 @@
...
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const [currentPage, setCurrentPage] = useState<number>(1);
+  const [perPage, setPerPage] = useState<number>(5);
```
Anastasiia-Svintsova, line 11: Just want you to know that TS will automatically infer the type based on the initial value. So you can simplify the code 
```suggestion
  const [currentPage, setCurrentPage] = useState(1);
  const [perPage, setPerPage] = useState(5);
```

src/App.tsx
```diff
@@ -1,18 +1,39 @@
...
+
+  const startIndex = (currentPage - 1) * perPage;
+  let endIndex = currentPage * perPage;
+
+  if (endIndex > items.length) {
+    endIndex = items.length;
+  }
```
Anastasiia-Svintsova, line 18: Maybe you can replace this code with a ternary operator?

src/App.tsx
```diff
@@ -21,6 +42,8 @@ export const App: React.FC = () => {
             data-cy="perPageSelector"
             id="perPageSelector"
             className="form-control"
+            onChange={(e) => selectPerPage(e)}
+            value={perPage}
           >
             <option value="3">3</option>
```
Anastasiia-Svintsova, line 48: Would be great to create an array with possible options and iterate over it, to not repeat the same code for each option

src/App.tsx
```diff
@@ -34,62 +57,19 @@ export const App: React.FC = () => {
...
-          </a>
-        </li>
-      </ul>
+      <Pagination
+        total={items.length}
+        perPage={perPage}
+        currentPage={currentPage}
+        onPageChange={(page) => {
+          setCurrentPage(page);
+        }}
```
Anastasiia-Svintsova, line 66: You can simplify it
```suggestion
        onPageChange={setCurrentPage}
```

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,94 @@
...
+              selectPage(e);
+            }}
+          >
+            {item}
+          </a>
+        </li>
+      ))}
+      <li
+        className={cn('page-item', {
+          disabled: currentPage === totalPageNumber,
```
Anastasiia-Svintsova, line 79: You use this `currentPage === totalPageNumber` value not once. Would be good to move it to a variable

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,94 @@
...
+    if (currentPage < totalPageNumber) {
+      onPageChange(currentPage + 1);
+    }
+  };
+
+  return (
+    <ul className="pagination">
+      <li
+        className={cn('page-item', {
+          disabled: currentPage === 1,
```
Anastasiia-Svintsova, line 45: Same here

Anastasiia-Svintsova (changes requested):

src/App.tsx
denis-kononenko: Yeah, sure, I forgot) I will fix

src/App.tsx
denis-kononenko: Yeah I can, but I saw decision with Math.min and I liked it more)

src/App.tsx
denis-kononenko: ok, got it, thanks

src/App.tsx
denis-kononenko: thank you for a such detailed CR 🙌

volodymyr-soltys97 (approved):
Well done 🔥

PR #547, opened 2023-07-21
react_pagination

vintwp:
https://vintwp.github.io/react_pagination/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,51 @@
-import React from 'react';
+import React, { useMemo, useState } from 'react';
 import './App.css';
-import { getNumbers } from './utils';
+import { getNumbers, getMaxNumberPage } from './utils';
+import { Pagination } from './components/Pagination';
+
+const PAGE_INITIAL = 1;
+const ITEMS_PER_PAGE_INITIAL = 5;
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const itemsList = useMemo(() => {
+    return [...items];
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -21,11 +54,14 @@ export const App: React.FC = () => {
             data-cy="perPageSelector"
             id="perPageSelector"
             className="form-control"
+            defaultValue={ITEMS_PER_PAGE_INITIAL}
+            onChange={e => handerItemsPerPage(e)}
           >
             <option value="3">3</option>
```
Anastasiia-Svintsova, line 60: Would be great to create an array with possible options and iterate over it, to not repeat the same code for each option

src/App.tsx
```diff
@@ -21,11 +54,14 @@ export const App: React.FC = () => {
...
             <option value="5">5</option>
             <option value="10">10</option>
             <option value="20">20</option>
+            <option value={itemsQty}>View all</option>
```
Anastasiia-Svintsova, line 64: The `View all` option doesn't exist in the task

src/App.tsx
```diff
@@ -34,62 +70,23 @@ export const App: React.FC = () => {
...
-          </a>
-        </li>
-      </ul>
+      <Pagination
+        total={itemsQty}
+        perPage={itemsPerPage}
+        currentPage={page}
+        onPageChange={(newPageNumber) => {
+          setPage(newPageNumber);
+        }}
```
Anastasiia-Svintsova, line 79: You can simplify it 
```suggestion
        onPageChange={setPage}
```

Anastasiia-Svintsova (changes requested):

volodymyr-soltys97 (approved):
Nice work 🔥

PR #548, opened 2023-07-22
done

vetal-hovenko:
[DEMO LINK](https://vetal-hovenko.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,44 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
+import { Pagination } from './components/Pagination';
 import { getNumbers } from './utils';
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
+const items: string[] = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const [amountOfItems, setAmountOfItems] = useState(5);
+  const [currentPage, setCurrentPage] = useState(0);
+  const [info, setInfo] = useState(`Page 1 (items 1 - 5 of ${items.length})`);
+
+  const startItem: number = currentPage * amountOfItems;
+  const pages: number[] = [];
+
+  const newList: string[] = items.slice(startItem, startItem + amountOfItems);
+
+  for (let i = 1; i <= Math.ceil(items.length / amountOfItems); i += 1) {
+    pages.push(i);
+  }
+
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+        {info}
       </p>
 
       <div className="form-group row">
         <div className="col-3 col-sm-2 col-xl-1">
           <select
+            value={amountOfItems}
             data-cy="perPageSelector"
             id="perPageSelector"
             className="form-control"
+            onChange={(event) => {
+              setCurrentPage(0);
+              setAmountOfItems(+event.target.value);
+            }}
           >
             <option value="3">3</option>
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,26 +1,44 @@
...
 import { getNumbers } from './utils';
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
+const items: string[] = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const [amountOfItems, setAmountOfItems] = useState(5);
+  const [currentPage, setCurrentPage] = useState(0);
```
volodymyr-soltys97, line 11: The currentPage state is initialized to 0, but it should be 1 as per the requirements. It says 'currentPage' (optional, defaults to 1).

src/App.tsx
```diff
@@ -1,26 +1,44 @@
...
+  const [info, setInfo] = useState(`Page 1 (items 1 - 5 of ${items.length})`);
+
+  const startItem: number = currentPage * amountOfItems;
+  const pages: number[] = [];
+
+  const newList: string[] = items.slice(startItem, startItem + amountOfItems);
+
+  for (let i = 1; i <= Math.ceil(items.length / amountOfItems); i += 1) {
```
volodymyr-soltys97, line 19: Avoid using a `for` loop to create a new array with each element increased by one. Instead, use `map`.

src/App.tsx
```diff
@@ -1,26 +1,44 @@
...
       </p>
 
       <div className="form-group row">
         <div className="col-3 col-sm-2 col-xl-1">
           <select
+            value={amountOfItems}
             data-cy="perPageSelector"
             id="perPageSelector"
             className="form-control"
+            onChange={(event) => {
```
volodymyr-soltys97, line 38: When the number of items per page is changed, the currentPage should remain the same, not reset to 0. The user might change the items per page while being on any page, not necessarily the first page.

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,95 @@
-export const Pagination = () => {};
+import { useEffect } from 'react';
+
+interface Props {
```
volodymyr-soltys97, line 3: The 'pages' prop is not defined in the task. It seems like this prop is used to store the possible pages. The component should calculate this dynamically based on the 'total', 'perPage' and 'currentPage' props.

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,95 @@
...
+            className="page-link"
+            href="#prev"
+            aria-disabled={isFirstPage}
+            onClick={goToPrevPage}
+          >
+            «
+          </a>
+        </li>
+
+        {pages.map(page => (
```
volodymyr-soltys97, line 61: The current page is highlighted by comparing indexes, which might be inefficient. Instead, directly compare the 'page' with 'currentPage'.

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,95 @@
...
+  setCurrentPage,
+  setInfo,
+  items,
+}) => {
+  const indexOfFirst = items.indexOf(newList[0]) + 1;
+  const indexOfLast = items.indexOf(newList[newList.length - 1]) + 1;
+
+  useEffect(() => {
+    setInfo(
+      `Page ${currentPage + 1} (items ${indexOfFirst} - ${
```
volodymyr-soltys97, line 25: Use template literals to compose strings. It's easier to read and understand.

volodymyr-soltys97 (changes requested):
Good job 👍 

volodymyr-soltys97 (approved):

PR #550, opened 2023-07-23
solution

Kinderwagen:
[DEMO LINK](https://Kinderwagen.github.io/react_pagination/)

### DIFF
--- a/src/components/Pagination/Pagination.tsx
+++ b/src/components/Pagination/Pagination.tsx
@@ -1,1 +1,96 @@
-export const Pagination = () => {};
+import React from 'react';
+import cn from 'classnames';
+import { getNumbers } from '../../utils';
+
+type Props = {
+  total: number,
+  perPage: number,
+  currentPage: number,
+  onPageChange: (page: number) => void,
+};
+
+export const Pagination: React.FC<Props> = ({
+  total,
+  perPage,
+  currentPage,
+  onPageChange,
+}) => {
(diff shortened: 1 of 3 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,96 @@
...
+          data-cy="prevLink"
+          className="page-link"
+          href="#prev"
+          aria-disabled={isFirstPage}
+          onClick={handleClickPrevLink}
+        >
+          «
+        </a>
+      </li>
+      {pageNumbers.map(item => (
```
VitaliyBodnar, line 60: try to provide more descriptive name than  just 'item'

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,96 @@
...
+            active: item === currentPage,
+          })}
+        >
+          <a
+            data-cy="pageLink"
+            className="page-link"
+            href="#1"
+            onClick={(e) => {
+              selectPage(e);
+            }}
```
VitaliyBodnar, line 73: in this case, using an event is unnecessary, because you can use a page from the pageNumbers.map
```suggestion
            onClick={() => selectPage(item)}
```

VitaliyBodnar (changes requested):
Almost done!

VitaliyBondarenko1982 (approved):

PR #551, opened 2023-07-23
react pagination

artyomwhite:
[DEMO LINK](https://artyomwhite.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,36 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const [perPage, setPerPage] = useState(5);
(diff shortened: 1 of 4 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,104 @@
...
+  onPageChange,
+}) => {
+  const pagesCount = Math.ceil(total / perPage);
+  const pages = [];
+  const isFirstPage = currentPage === 1;
+  const isLastPage = currentPage === pagesCount;
+
+  for (let i = 1; i <= pagesCount; i += 1) {
+    pages.push(i);
+  }
```
VitaliyBodnar, line 26: it's better to move this code to function as it will a code cleaner

src/App.tsx
```diff
@@ -34,63 +54,17 @@ export const App: React.FC = () => {
...
-      </ul>
+      <Pagination
+        items={visibleItems}
+        total={itemsSize}
+        perPage={perPage}
+        currentPage={currentPage}
+        onPageChange={(page) => {
+          if (page !== currentPage) {
+            setCurrentPage(page);
+          }
```
VitaliyBodnar, line 65: try to not use a lot of logic in a template, better to move this code to the separate function

VitaliyBodnar (changes requested):
Almost done!

PR #557, opened 2023-07-26
Solution

Viktoriia6666:
[DEMO LINK](https://Viktoriia6666.github.io/react_pagination/)

### DIFF
--- a/src/Pagination.tsx
+++ b/src/Pagination.tsx
@@ -1,0 +1,128 @@
+import classNames from 'classnames';
+
+type Props = {
+  total: number;
+  perPage: number;
+  currentPage: number;
+  onPageChange: (value: number) => void;
+};
+
+const getNumberOfPages = (
+  total: number,
+  perPage: number,
+): number[] => {
+  const allPages: number[] = [];
+
+  for (let i = 1; i <= total; i += 1) {
+    allPages.push(i);
+  }
+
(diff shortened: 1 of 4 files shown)


src/Pagination.tsx
```diff
@@ -0,0 +1,128 @@
...
+  currentPage,
+  onPageChange,
+}) => {
+  const totalPages = getNumberOfPages(total, perPage);
+  const elementsOnPage = getNumberOfElements(currentPage, perPage, total);
+  const isFirstPage = currentPage === 1;
+  const isLastPage = currentPage === Math.max(...totalPages);
+
+  const onPageLinkClick = (page: number) => {
+    if (page !== currentPage) {
```
ivbutsykin, line 52: It's should be okay, if you remove this `if`. Because your page will be the same and react won't rerender components

src/Pagination.tsx
```diff
@@ -0,0 +1,128 @@
...
+  onPageChange: (value: number) => void;
+};
+
+const getNumberOfPages = (
+  total: number,
+  perPage: number,
+): number[] => {
+  const allPages: number[] = [];
+
+  for (let i = 1; i <= total; i += 1) {
```
ivbutsykin, line 16: Preferable to use array methods like `map`, `filter`, etc

ivbutsykin (approved):
Good job!

PR #571, opened 2023-08-10
pagination was introduced

w315:
[DEMO LINK](https://w315.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,96 +1,46 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
-import { getNumbers } from './utils';
+import { getNumbers, getPages } from './utils';
+import { Table } from './components/Table/Table';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
+export function getItems(page: number, limit: number): string[] {
+  const visibleItems = [];
+
+  for (let i = (page - 1) * limit; i < (page * limit); i += 1) {
+    visibleItems.push(items[i]);
+  }
+
+  return visibleItems;
+}
+
+export function getLength() {
+  return items.length;
+}
+
 export const App: React.FC = () => {
+  const [page, setPage] = useState<number>(1);
+  const [limit, setLimit] = useState<number>(5);
+
+  const totalPage = Math.ceil(getLength() / limit);
+
+  const pages = getPages(1, totalPage)
+    .map(n => `${n}`);
+
   return (
     <div className="container">
-      <h1>Items with Pagination</h1>
-
-      <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
-      </p>
-
-      <div className="form-group row">
-        <div className="col-3 col-sm-2 col-xl-1">
-          <select
-            data-cy="perPageSelector"
-            id="perPageSelector"
-            className="form-control"
-          >
-            <option value="3">3</option>
-            <option value="5">5</option>
-            <option value="10">10</option>
-            <option value="20">20</option>
-          </select>
-        </div>
-
-        <label htmlFor="perPageSelector" className="col-form-label col">
-          items per page
-        </label>
-      </div>
-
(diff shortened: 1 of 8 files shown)


src/utils.ts
```diff
@@ -7,3 +7,13 @@ export function getNumbers(from: number, to: number): number[] {
 
   return numbers;
 }
+
+export function getPages(from: number, to: number): number[] {
+  const numbers = [];
+
+  for (let n = from; n <= to; n += 1) {
```
volodymyr-soltys97, line 14: You also can use Array.from method to create an array with specified range. For example, const numbers = Array.from({length: to - from + 1}, (_, i) => from + i);

volodymyr-soltys97, line 14: Avoid using a 'for' loop to create a new array with modified values. Instead, use 'map'.

volodymyr-soltys97, line 14: Use 'map' to create a new array with modified values.

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,62 @@
...
+            }
+          }}
+        >
+          «
+        </a>
+      </li>
+      {
+        pages.map(page => {
+          return (
+            <li className={`page-item ${+page === pageCur ? 'active' : ''}`}>
```
volodymyr-soltys97, line 28: The 'li' elements created by the map function need a unique 'key' prop to avoid potential rendering issues.

src/components/Items/Items.tsx
```diff
@@ -0,0 +1,27 @@
...
+  view: () => string[];
+};
+
+export const Items: React.FC<Props> = ({
+  view,
+}) => {
+  const items = view();
+
+  return (
+    <ul>
```
volodymyr-soltys97, line 11: There is no check if the items array is not empty before rendering the list. If the items array is empty, an empty <ul> tag will be rendered. Consider adding a condition to only render the list if the items array is not empty.

src/components/Table/Table.tsx
```diff
@@ -0,0 +1,67 @@
...
+export const Table: React.FC<Props> = ({
+  view,
+  pages,
+  pageCur,
+  items,
+  limit,
+  setPage,
+  setLimit,
+}) => {
+  const startItem: number = pageCur * limit + 1 - limit;
```
volodymyr-soltys97, line 23: There's a potential off-by-one error here. If the 'pageCur' is 1, the 'startItem' will be 1 which is correct. But if the 'pageCur' is 2, the 'startItem' will be 'limit + 1' which may cause the first item of the second page to be skipped. The correct calculation should be '(pageCur - 1) * limit + 1'.

src/App.tsx
```diff
@@ -1,96 +1,46 @@
...
-  return (
-    <div className="container">
-      <h1>Items with Pagination</h1>
+export function getItems(page: number, limit: number): string[] {
+  const visibleItems = [];
 
-      <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
-      </p>
+  for (let i = (page - 1) * limit; i < (page * limit); i += 1) {
```
volodymyr-soltys97, line 13: You are using a for loop to create a new array. Consider using the map function for a cleaner solution. Also, this code will fail if the array length is less than page*limit. We need to add a condition to check if the items[i] exists before pushing it into visibleItems.

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,62 @@
...
+        })
+      }
+      <li className={`page-item ${pageCur !== pages.length ? '' : 'disabled'}`}>
+        <a
+          data-cy="nextLink"
+          className="page-link"
+          href="#next"
+          aria-disabled={pageCur === pages.length ? 'true' : 'false'}
+          onClick={() => {
+            if (pageCur !== pages.length) {
```
volodymyr-soltys97, line 52: Create a variable for this checking and use it everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

anastasiiavorobiova (approved):
Great work!

PR #572, opened 2023-08-10
solution 1.0

aliliolek:
[DEMO LINK](https://aliliolek.github.io/react_pagination/)

### DIFF
--- a/src/components/Pagination/Pagination.tsx
+++ b/src/components/Pagination/Pagination.tsx
@@ -1,1 +1,81 @@
-export const Pagination = () => {};
+import React from 'react';
+
+type Props = {
+  total: number,
+  // perPage: number,
+  currentPage: number,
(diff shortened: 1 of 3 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,81 @@
...
+          className="page-link"
+          href="#prev"
+          aria-disabled={disableGoToPrevious}
+          onClick={goToPreviousPage}
+        >
+          «
+        </a>
+      </li>
+
+      {Array.from({ length: numberOfPages }, (_, index) => index + 1)
```
maxim2310, line 47: would be nice to move it as separate func

maxim2310 (changes requested):
Almost there) Just one more thing needs to be fixed

witflash (approved):
Well done 👍

PR #575, opened 2023-08-14
solution

JulyaPetrovskaya:
[DEMO LINK](https://JulyaPetrovskaya.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,42 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
-
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
-  .map(n => `Item ${n}`);
+import { Pagination } from './components/Pagination/Pagination';
 
 export const App: React.FC = () => {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const items = getNumbers(1, 42).map((n) => `Item ${n}`);
+
+  const [current, setCurrent] = useState(1);
+  const [perPage, setPerPage] = useState(5);
+
+  const handlePageChange = (page: number) => {
+    setCurrent(page);
+  };
+
+  const handlePerPageChange = (e: React.ChangeEvent<HTMLSelectElement>) => {
+    setPerPage(Number(e.target.value));
+  };
+
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+        Page
+        {' '}
+        {current}
+        {' '}
+        (items
+        {' '}
+        {(current - 1) * perPage + 1}
+        {' '}
+        -
+        {Math.min(current * perPage, items.length)}
+        {' '}
+        of
+        {items.length}
+        )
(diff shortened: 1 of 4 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,87 @@
...
+  total,
+  perPage,
+  currentPage = 1,
+  onPageChange,
+}) => {
+  const totalPages = Math.ceil(total / perPage);
+
+  const pageNumbers: number[] = [];
+
+  for (let i = 1; i <= totalPages; i += 1) {
```
Viktor-Kost, line 20: Avoid using a `for` loop to create a new array with each element increased by one. Instead, use `map`.

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,87 @@
...
+          onClick={handleNextClick}
+        >
+          »
+        </a>
+      </li>
+    </ul>
+  );
+};
+
+export default Pagination;
```
Viktor-Kost, line 87: The component is exported twice, first as a named export on line 11 and then as a default export on line 88. It's not necessarily wrong but it may lead to confusion. Stick to one type of export per component to maintain consistency.

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,87 @@
...
+    }
+  };
+
+  return (
+    <ul className="pagination">
+      <li className={`page-item ${currentPage === 1 ? 'disabled' : ''}`}>
+        <a
+          data-cy="prevLink"
+          className="page-link"
+          href="#prev"
```
Viktor-Kost, line 44: Please check your deploy, since whenever I click at a number of page or prev/next button the link isn't changing

src/App.tsx
```diff
@@ -34,62 +51,19 @@ export const App: React.FC = () => {
...
+        onPageChange={handlePageChange}
+      />
+
       <ul>
-        <li data-cy="item">Item 1</li>
-        <li data-cy="item">Item 2</li>
-        <li data-cy="item">Item 3</li>
-        <li data-cy="item">Item 4</li>
-        <li data-cy="item">Item 5</li>
+        {items.slice((current - 1) * perPage, current * perPage).map((item) => (
```
Viktor-Kost, line 62: It's better to move this calculation to a variable for better readability

src/App.tsx
```diff
@@ -1,18 +1,33 @@
...
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+        {`Page ${current} (items ${(current - 1) * perPage + 1} - ${Math.min(
+          current * perPage,
+          items.length,
+        )} of ${items.length})`}
```
Viktor-Kost, line 30: It's better to move this calculation to a variable for better readability

Viktor-Kost (changes requested):
Nice!
Some moments can be improved

src/components/Pagination/Pagination.tsx
JulyaPetrovskaya: what do i need to do to fix this?

VitaliyBondarenko1982 (approved):
Well done

PR #590, opened 2023-08-27
add solution

purpleFade:
[DEMO LINK](https://purplefade.github.io/react_pagination/)

### DIFF
--- a/src/components/Pagination/Pagination.tsx
+++ b/src/components/Pagination/Pagination.tsx
@@ -1,1 +1,87 @@
-export const Pagination = () => {};
+import { getNumbers, getNumbersInRange } from '../../utils';
+
+type Props = {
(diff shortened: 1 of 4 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,87 @@
...
+    1,
+    total,
+  );
+
+  return (
+    <>
+      {/* Pagination links */}
+      <ul className="pagination">
+        {/* Previous page link */}
+        <li className={`page-item ${currentPage === 1 ? 'disabled' : ''}`}>
```
SanyaBratashchuk, line 39: it is better to use libraries like classnames or clsx for conditional classNames

SanyaBratashchuk (approved):

PR #594, opened 2023-08-29
pagination solution

AnVBondar:
[DEMO LINK](https://AnVBondar.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,64 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
+function getItems(
+  listOfItems: string[],
+  itemNumbers: string,
+): string[][] {
+  const result = [];
+  let itemsPart = [];
+  let count = 0;
+
+  for (let i = 0; i < listOfItems.length; i += 1) {
+    if (i === listOfItems.length - 1) {
+      itemsPart.push(listOfItems[i]);
+      result.push(itemsPart);
+      break;
+    }
+
+    itemsPart.push(listOfItems[i]);
+    count += 1;
+
+    if (count === +itemNumbers) {
+      result.push(itemsPart);
+      itemsPart = [];
+      count = 0;
+    }
+  }
+
+  return result;
+}
+
 export const App: React.FC = () => {
+  const [perPage, setPerPage] = useState('5');
+  const [currentPage, setCurrentPage] = useState(1);
+  const groupedItems = getItems(items, perPage);
+  const currentGroupOfItems = groupedItems[currentPage - 1];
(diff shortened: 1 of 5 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,91 @@
...
+        'page-item',
+        { disabled: isFirstPage() },
+      )}
+      >
+        <a
+          data-cy="prevLink"
+          className="page-link"
+          href="#prev"
+          aria-disabled={isFirstPage()}
+          onClick={() => !isFirstPage() && onPageChange(currentPage - 1)}
```
Viktor-Kost, line 51: no needs to check lat page at onClick since you disable this button

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,91 @@
...
+    return false;
+  };
+
+  const isLastPage = (): boolean => {
+    if (currentPage === totalPages) {
+      return true;
+    }
+
+    return false;
+  };
```
Viktor-Kost, line 37: The function 'isLastPage' can be simplified to return the condition directly without the 'if' statement.
```suggestion
  const isLastPage = currentPage === totalPages;
```

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,91 @@
...
+    numberOfPages.push(i);
+  }
+
+  const isFirstPage = (): boolean => {
+    if (currentPage === 1) {
+      return true;
+    }
+
+    return false;
+  };
```
Viktor-Kost, line 29: The function 'isFirstPage' can be simplified to return the condition directly without the 'if' statement.
```suggestion
  const isFirstPage = currentPage === 1;
```

src/App.tsx
```diff
@@ -1,18 +1,64 @@
...
+      itemsPart = [];
+      count = 0;
+    }
+  }
+
+  return result;
+}
+
 export const App: React.FC = () => {
+  const [perPage, setPerPage] = useState('5');
```
Viktor-Kost, line 39: The 'perPage' prop should be of type 'number'. This is because it is used in arithmetic operations, and it is better to avoid type coercion.
```suggestion
  const [perPage, setPerPage] = useState(5);
```

Viktor-Kost (changes requested):

src/components/Pagination/Pagination.tsx
AnVBondar: originally I had code like this `onClick={() => onPageChange(currentPage - 1)}` it was enough as button disabled but due to tests failed I had to add condition.

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,91 @@
...
+        'page-item',
+        { disabled: isLastPage() },
+      )}
+      >
+        <a
+          data-cy="nextLink"
+          className="page-link"
+          href="#next"
+          aria-disabled={isLastPage()}
+          onClick={() => !isLastPage() && onPageChange(currentPage + 1)}
```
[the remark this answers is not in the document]
AnVBondar: the same reason as in previous comment.

volodymyr-soltys97 (approved):

PR #598, opened 2023-09-03
solution

drhtka:
https://drhtka.github.io/react_pagination/


etojeDenys (changes requested):
let's fix this case firstly
image
should be 3 items per page

src/App.tsx
```diff
@@ -1,18 +1,36 @@
...
   .map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const [page, setPage] = useState(1);
+  const [itemsPerPage, setItemsPerPage] = useState(5);
+
+  const firstItem = (page - 1) * itemsPerPage;
+  const lastItem = page * itemsPerPage < items.length
+    ? page * itemsPerPage
+    : items.length;
```
etojeDenys, line 17: you can just use `Math.max(...)`

etojeDenys (changes requested):
nit:

PR #611, opened 2023-09-06
Develop

domkab:
https://domkab.github.io/react_pagination


anastasiia-tilikina (changes requested):
your solution for working with url is a bit inappropriate.
your should use [react router](https://reactrouter.com/en/main) and [useSearchParams](https://reactrouter.com/en/main/hooks/use-search-params) hook. Also consider implementing simplified version of this task since you will learn these topics later)

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,105 @@
...
+              data-cy={`page-${page}`}
+            >
+              <a
+                className="page-link"
+                data-cy="pageLink"
+                href={`#${page}`}
+                onClick={(e) => {
+                  e.preventDefault();
+                  handlePageClick(page);
+                }}
```
anastasiia-tilikina, line 79:
```suggestion
                onClick={() => handlePageClick(page)}
```
Basically, you don't need to prevent default behavior on link clicks. let's remove it every where

anastasiia-tilikina (changes requested):
Well done! 
Just one small fix left.

BTW you did a great job in the previous iteration with implementing more complicated logic) There is just easier way you will learn later)

Anastasiia-Svintsova (approved):

PR #614, opened 2023-09-06
Solution

Serhii-Shtokalo:
- [DEMO LINK](https://Serhii-Shtokalo.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,35 @@
-import React from 'react';
+import React, { ChangeEvent, useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
-
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
-  .map(n => `Item ${n}`);
+import { Pagination } from './components/Pagination';
 
 export const App: React.FC = () => {
+  const totalItems = 42;
+  const firstPage = 1;
+
+  const [itemsOnPage, setItemsOnPage] = useState(5);
+  const [initialPage, setInitialPage] = useState(firstPage);
+
+  const maxItems = initialPage * itemsOnPage;
+  const startVisibleItems = maxItems - itemsOnPage;
+  const endVisibleItems = maxItems > totalItems ? totalItems : maxItems;
+
+  const selectedItems = (event: ChangeEvent<HTMLSelectElement>) => {
+    setItemsOnPage(+event.target.value);
+  };
+
+  const changedPage = (page: number) => {
+    setInitialPage(page);
+  };
+
   return (
     <div className="container">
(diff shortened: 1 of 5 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,97 @@
...
+      <li
+        className={cn('page-item', {
+          disabled: chosenPage === totalTabs,
+        })}
+      >
+        <a
+          data-cy="nextLink"
+          className="page-link"
+          href="#next"
+          aria-disabled={chosenPage === totalTabs ? 'true' : 'false'}
```
etojeDenys, line 89:
```suggestion
          aria-disabled={chosenPage === totalTabs}
```

etojeDenys: better to move this condition to some var and reuse everywhere

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,97 @@
...
+  onPageChange: (page: number) => void;
+}
+
+export const Pagination: FC<Props> = ({
+  total,
+  perPage,
+  currentPage,
+  onPageChange,
+}) => {
+  const [chosenPage, setChosenPage] = useState(currentPage);
```
etojeDenys, line 18: why do we need this useState if we can directly call onPageChange?

etojeDenys (changes requested):

src/components/Pagination/Pagination.tsx
Viktor-Kost: isn't fixed. You shouldn't use an additional state, you can just use onPageChange function directly at Pagination.tsx

Viktor-Kost (changes requested):
Please check this mentor's comment https://github.com/mate-academy/react_pagination/pull/614#discussion_r1318140204

If you need any help feel free to ask at fe_chat

src/App.tsx
```diff
@@ -1,18 +1,26 @@
...
+
+const items = getNumbers(1, total).map((n) => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const [currentPage, setCurrentPage] = useState<number>(1);
+  const [perPage, setPerPage] = useState<number>(5);
+
+  const START_ITEM = perPage * currentPage - perPage + 1;
+  const END_ITEM =
+    perPage * currentPage > items.length ? items.length : perPage * currentPage;
```
TarasHoliuk, line 16:
```suggestion
  const END_ITEM = perPage * currentPage > items.length
    ? items.length 
    : perPage * currentPage;
```
to fix the lint error - FYI

TarasHoliuk (approved):

PR #621, opened 2023-09-10
implement Pagination component

KatOlista:
[DEMO LINK](https://KatOlista.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,32 +1,65 @@
-import React from 'react';
+import React, { useState } from 'react';
+
+// import { createBrowserRouter } from 'react-router-dom';
+
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
+import { ItemList } from './components/ItemList';
+import { PerPageList } from './components/PerPageList';
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
-  .map(n => `Item ${n}`);
+const defaultPaginVal = {
+  total: 42,
+  perPage: 5,
+  currentPage: 1,
+};
+
+const options = [3, 5, 10, 20];
 
 export const App: React.FC = () => {
+  const [paginationOption, setPaginationOption] = useState({
+    total: defaultPaginVal.total,
+    perPage: defaultPaginVal.perPage,
+    currentPage: defaultPaginVal.currentPage,
+  });
+
+  const itemPerPageHandler = (value: string) => setPaginationOption({
+    ...paginationOption,
+    ...{ perPage: +value, currentPage: defaultPaginVal.currentPage },
+  });
+
+  const currentPageHandler = (value: number) => setPaginationOption({
+    ...paginationOption, ...{ currentPage: value },
+  });
+
+  const fromItem = (paginationOption?.currentPage - 1)
+    * paginationOption.perPage + 1;
+
+  const maxCountItem = paginationOption?.currentPage * paginationOption.perPage;
+
+  const toItem = maxCountItem
+    > defaultPaginVal.total
+    ? defaultPaginVal.total
+    : maxCountItem;
+
+  const items = getNumbers(fromItem, toItem)
+    .map(n => `Item ${n}`);
+
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+        {`Page ${paginationOption.currentPage} (items ${fromItem} - ${toItem} of ${paginationOption.total})`}
       </p>
 
       <div className="form-group row">
         <div className="col-3 col-sm-2 col-xl-1">
-          <select
-            data-cy="perPageSelector"
-            id="perPageSelector"
-            className="form-control"
-          >
-            <option value="3">3</option>
-            <option value="5">5</option>
-            <option value="10">10</option>
-            <option value="20">20</option>
-          </select>
+          <PerPageList
+            selectOptions={options}
+            paginationOption={paginationOption}
+            onChangePaginationOption={itemPerPageHandler}
+          />
         </div>
 
         <label htmlFor="perPageSelector" className="col-form-label col">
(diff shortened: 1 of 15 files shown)


src/App.tsx
```diff
@@ -1,96 +1,78 @@
-import React from 'react';
+import React, { useState } from 'react';
+
+// import { createBrowserRouter } from 'react-router-dom';
```
roman-mirzoian, line 3: Do not leave unnecessary comments.

src/App.tsx
```diff
@@ -1,96 +1,78 @@
...
+  });
+
+  const itemPerPageHandler = (value: string) => setPaginationOption({
+    ...paginationOption,
+    ...{ perPage: +value, currentPage: defaultPaginVal.currentPage },
+  });
+
+  const currentPageHandler = (value: number) => setPaginationOption({
+    ...paginationOption, ...{ currentPage: value },
+  });
```
roman-mirzoian, line 33: If we want to work with the actual state, it should be received through a callback, and not applied directly.
It is also not necessary to wrap the field in an object and immediately open it if it is not necessary, just overwrite the required fields:
```js
const oldObject = {
  field1: 1,
  field2: 2
};
const newObject = {
  ...oldObject,
  field1: 2
};
const object = {
  ...newObject,
  field2: 1
};
```

src/App.tsx
```diff
@@ -1,96 +1,78 @@
...
+
+  const fromItem = (paginationOption?.currentPage - 1)
+    * paginationOption.perPage + 1;
+
+  const maxCountItem = paginationOption?.currentPage * paginationOption.perPage;
+
+  const toItem = maxCountItem
+    > defaultPaginVal.total
+    ? defaultPaginVal.total
+    : maxCountItem;
```
roman-mirzoian, line 43:
```suggestion
  const toItem = Math.min(maxCountItem, defaultPaginVal.total);
```

src/App.tsx
```diff
@@ -1,96 +1,78 @@
...
+
+  const items = getNumbers(fromItem, toItem)
+    .map(n => `Item ${n}`);
```
roman-mirzoian, line 46: This function should be moved outside the component.

src/components/PerPageList/PerPageList.tsx
```diff
@@ -0,0 +1,32 @@
...
+      data-cy="perPageSelector"
+      id="perPageSelector"
+      className="form-control"
+      onChange={(event) => onChangePaginationOption(event.target.value)}
+    >
+      {selectOptions.map(option => (
+        <PerPageOption
+          key={option}
+          option={option}
+        />
```
roman-mirzoian, line 28: There is no need to create a separate component for each option tag, it is better to keep this as one component with the select tag.

src/components/PerPageList/PerPageList.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { PerPageOption } from '../PerPageOption';
+import { PaginationOption } from '../../types/PaginationOption';
+
+type Props = {
+  selectOptions: number[],
+  paginationOption: PaginationOption,
+  onChangePaginationOption: (value: string) => void,
+};
+
+export const PerPageList: React.FC<Props> = ({
```
roman-mirzoian, line 11:
```suggestion
export const PageSizeSelectList: React.FC<Props> = ({
```

src/components/PerPageList/PerPageList.tsx
```diff
@@ -0,0 +1,32 @@
...
+  paginationOption,
+  onChangePaginationOption,
+}) => {
+  return (
+    <select
+      value={paginationOption.perPage}
+      data-cy="perPageSelector"
+      id="perPageSelector"
+      className="form-control"
+      onChange={(event) => onChangePaginationOption(event.target.value)}
```
roman-mirzoian, line 22: It is better to handle the event entirely inside the `onChangePaginationOption` function
```suggestion
      onChange={onChangePaginationOption}
```

src/App.tsx
```diff
@@ -1,96 +1,78 @@
...
+const defaultPaginVal = {
+  total: 42,
+  perPage: 5,
+  currentPage: 1,
+};
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
-  .map(n => `Item ${n}`);
+const options = [3, 5, 10, 20];
```
roman-mirzoian, line 17:
```suggestion
const PAGE_SIZE_OPTIONS = [3, 5, 10, 20];
```

roman-mirzoian (changes requested):
Good job, but there is quite a bit of work to be done.

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,84 @@
...
+  const onPrevPageHandler = () => {
+    if (paginationOption.currentPage > 1) {
+      onPageChange(startVal - 1);
+    }
+  };
+
+  return (
+    <ul
+      className="pagination"
+    >
```
roman-mirzoian, line 41:
```suggestion
    <ul className="pagination">
```

src/utils.ts
```diff
@@ -1,9 +1,5 @@
-export function getNumbers(from: number, to: number): number[] {
-  const numbers = [];
-
-  for (let n = from; n <= to; n += 1) {
-    numbers.push(n);
-  }
-
-  return numbers;
-}
+export const defaultPaginVal = {
```
roman-mirzoian, line 1: Do not abbreviate names to unreadable ones.

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,84 @@
...
+
+type Props = {
+  paginationOption: PaginationOption,
+  onPageChange: (value: number) => void,
+};
+
+export const Pagination: React.FC<Props> = ({
+  paginationOption,
+  onPageChange,
+}) => {
```
roman-mirzoian, line 16: Fields with paginationOption are often reused, so it is worth destructuring this object.

src/components/ItemInfo/ItemInfo.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+type Props = {
+  item: string,
+};
+
+export const ItemInfo: React.FC<Props> = ({ item }) => {
```
roman-mirzoian, line 7: In this situation, such a division into components is a little inappropriate, since the removed component takes up much more space than dividing the tag into a straight line. It is worth separating complex components or components that are lexically different from each other, but if we have a list with simple tags, it is worth keeping everything together.

src/App.tsx
```diff
@@ -1,96 +1,63 @@
...
+  const [paginationOption, setPaginationOption] = useState({
+    total: defaultPaginVal.total,
+    perPage: defaultPaginVal.perPage,
+    currentPage: defaultPaginVal.currentPage,
+  });
+
+  const currentPageHandler = (value: number) => setPaginationOption({
+    ...paginationOption,
+    currentPage: value,
+  });
```
roman-mirzoian, line 22: If we want to work with the actual state, it should be received through a callback, and not applied directly.

src/components/PageSizeSelectList/PageSizeSelectList.tsx
```diff
@@ -0,0 +1,38 @@
...
+};
+
+export const PageSizeSelectList: React.FC<Props> = ({
+  selectOptions,
+  paginationOption,
+  onSetPaginationOption,
+}) => {
+  const handleItemPerPage
+  = (event: React.ChangeEvent<HTMLSelectElement>): void => {
+    onSetPaginationOption({
```
roman-mirzoian, line 18: The same about actual state.

roman-mirzoian (changes requested):
Almost done!

voievodik (approved):
Good job!

src/components/PageSizeSelectList/PageSizeSelectList.tsx
roman-mirzoian: Not fixed 😢 
If we want to work with the actual state, it should be received through a callback, and not applied directly.

roman-mirzoian (changes requested):
One bug has not been fixed.

BudnikOleksii (approved):
Good work 👍

PR #622, opened 2023-09-11
Develop

EvhenIvchenko:
[DEMO LINK](https://EvhenIvchenko.github.io/react_pagination/)

### DIFF
--- a/src/components/PageSizeSelect/PageSizeSelect.tsx
+++ b/src/components/PageSizeSelect/PageSizeSelect.tsx
@@ -1,0 +1,31 @@
+enum SelectValue {
+  THREE = 3,
+  FIVE = 5,
+  TEN = 10,
+  TWENTY = 20,
+}
+
+type Props = {
+  perPage: number
+  onChangeSize: (event: React.ChangeEvent<HTMLSelectElement>) => void
+};
+
+export const PageSizeSelect: React.FC<Props> = ({
+  perPage,
+  onChangeSize,
+}) => {
+  return (
+    <select
+      data-cy="perPageSelector"
+      id="perPageSelector"
+      className="form-control"
+      defaultValue={perPage}
+      onChange={(onChangeSize)}
+    >
+      <option value={SelectValue.THREE}>{SelectValue.THREE}</option>
+      <option value={SelectValue.FIVE}>{SelectValue.FIVE}</option>
+      <option value={SelectValue.TEN}>{SelectValue.TEN}</option>
+      <option value={SelectValue.TWENTY}>{SelectValue.TWENTY}</option>
+    </select>
+  );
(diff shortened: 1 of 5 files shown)


src/components/PageSizeSelect/PageSizeSelect.tsx
```diff
@@ -0,0 +1,31 @@
...
+  perPage,
+  onChangeSize,
+}) => {
+  return (
+    <select
+      data-cy="perPageSelector"
+      id="perPageSelector"
+      className="form-control"
+      defaultValue={perPage}
+      onChange={(onChangeSize)}
```
lerastarynets, line 23: You should pass the onChange function directly to the onChange prop, instead of wrapping it in an anonymous function.

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,84 @@
-export const Pagination = () => {};
+import cn from 'classnames';
+
+type Props = {
+  total: number,
+  perPage : number,
+  currentPage: number,
```
lerastarynets, line 6: The prop 'currentPage' should be optional with a default value of 1 as required by the task.

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,84 @@
...
+          <a
+            data-cy="prevLink"
+            className="page-link"
+            href="#prev"
+            aria-disabled={currentPage === 1 ? 'true' : 'false'}
+            onClick={() => {
+              if (currentPage !== 1) {
+                onPageChange(currentPage - 1);
+              }
+            }}
```
lerastarynets, line 34: Pls move the function inside `onClick` handler to a separate const and use as a callback here

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,84 @@
...
+              'page-item',
+              { disabled: currentPage === pagesNumbers },
+            )
+          }
+        >
+          <a
+            data-cy="nextLink"
+            className="page-link"
+            href="#next"
+            aria-disabled={currentPage === pagesNumbers ? 'true' : 'false'}
```
lerastarynets, line 71: The 'aria-disabled' attribute should be a boolean, not a string. You need to remove the single quotes.

lerastarynets (changes requested):
Nice start! Let's fix the solution according to the checklist

src/components/PageSizeSelect/PageSizeSelect.tsx
sTorba24, line 23:
```suggestion
      onChange={onChangeSize}
```

sTorba24 (approved):

PR #637, opened 2023-09-20
Implement_react_pagination

rialleons:
- [DEMO LINK](https://rialleons.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,27 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const [perPage, setPerPage] = useState(5);
+  const [currentPage, setCurrentPage] = useState(1);
+
+  const itemsTotal = items.length;
+  const fromItem = (currentPage - 1) * perPage + 1;
+  const toItem = Math.min(currentPage * perPage, itemsTotal);
+
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+
+        {`Page ${currentPage} (items ${fromItem} - ${toItem} of ${itemsTotal})`}
       </p>
 
       <div className="form-group row">
@@ -21,6 +30,11 @@ export const App: React.FC = () => {
             data-cy="perPageSelector"
             id="perPageSelector"
             className="form-control"
+            value={perPage}
+            onChange={event => {
+              setPerPage(Number(event.target.value));
+              setCurrentPage(1);
+            }}
           >
             <option value="3">3</option>
             <option value="5">5</option>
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -34,62 +48,19 @@ export const App: React.FC = () => {
...
-          >
-            »
-          </a>
-        </li>
-      </ul>
+      <Pagination
+        total={itemsTotal}
+        perPage={perPage}
+        currentPage={currentPage}
+        onPageChange={(pageNumber) => setCurrentPage(pageNumber)}
```
etojeDenys, line 55:
```suggestion
        onPageChange={setCurrentPage}
```
we can simplify it

src/App.tsx
```diff
@@ -34,62 +48,19 @@ export const App: React.FC = () => {
...
+      />
+
       <ul>
-        <li data-cy="item">Item 1</li>
-        <li data-cy="item">Item 2</li>
-        <li data-cy="item">Item 3</li>
-        <li data-cy="item">Item 4</li>
-        <li data-cy="item">Item 5</li>
+        {items.map((item, index) => (
```
etojeDenys, line 59: i think we can just call getNumbers instead of this strange condition

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,77 @@
...
+  total,
+  perPage,
+  currentPage,
+  onPageChange,
+}) => {
+  const numberOfPages = Math.ceil(total / perPage);
+
+  const pageItems = Array.from(
+    { length: numberOfPages }, (_, index) => index + 1,
+  );
```
etojeDenys, line 20: same here, we can just call getNumbers

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,77 @@
...
+
+  const isFirstPage = currentPage === 1;
+  const isLastPage = currentPage === numberOfPages;
+
+  return (
+    <ul className="pagination">
+      <li className={cn({
+        'page-item': true,
+        disabled: isFirstPage,
+      })}
```
etojeDenys, line 30:
```suggestion
      <li className={cn('page-item' , {
        disabled: isFirstPage,
      })}
```

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -21,6 +30,11 @@ export const App: React.FC = () => {
             data-cy="perPageSelector"
             id="perPageSelector"
             className="form-control"
+            value={perPage}
+            onChange={event => {
+              setPerPage(Number(event.target.value));
+              setCurrentPage(1);
```
VitaliyBodnar, line 36: it is bad practise to make anonymous arrow functions in jsx.
make handler in component and pass reference here

src/App.tsx
```diff
@@ -34,62 +48,17 @@ export const App: React.FC = () => {
...
+        onPageChange={setCurrentPage}
+      />
+
       <ul>
-        <li data-cy="item">Item 1</li>
-        <li data-cy="item">Item 2</li>
-        <li data-cy="item">Item 3</li>
-        <li data-cy="item">Item 4</li>
-        <li data-cy="item">Item 5</li>
+        {items.slice(fromItem - 1, toItem).map(item => (
```
VitaliyBodnar, line 59: please don't call a function from the template because it will be executed every time when the state change, so make sense to move it to the variable

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,74 @@
...
+        >
+          «
+        </a>
+      </li>
+
+      {pageItems.map(pageItem => (
+        <li className={cn({
+          'page-item': true,
+          active: pageItem === currentPage,
+        })}
```
VitaliyBodnar, line 45: missed key attribute

VitaliyBodnar (changes requested):

VitaliyBodnar (approved):

PR #638, opened 2023-09-20
react-pagination

vlad-fil:
[DEMO LINK](https://vlad-fil.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,34 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const total = items.length;
+
+  const [currentPage, setCurrentPage] = useState(1);
+  const [perPage, setPerPage] = useState('5');
+
+  const onPageChange = (page: number) => {
+    setCurrentPage(page);
+  };
+
+  const lastPage = Math.ceil(total / Number(perPage));
+  const start = (currentPage - 1) * Number(perPage);
+  const end = (start + Number(perPage)) < total
+    ? start + Number(perPage)
+    : total;
+
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,18 +1,34 @@
...
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const total = items.length;
+
+  const [currentPage, setCurrentPage] = useState(1);
+  const [perPage, setPerPage] = useState('5');
```
etojeDenys, line 14:
```suggestion
  const [perPage, setPerPage] = useState(5);
```
it can be the number

src/App.tsx
```diff
@@ -21,6 +37,11 @@ export const App: React.FC = () => {
             data-cy="perPageSelector"
             id="perPageSelector"
             className="form-control"
+            defaultValue={5}
+            onChange={(event) => {
+              setPerPage(event.target.value);
```
etojeDenys, line 42:
```suggestion
              setPerPage(+event.target.value);
```

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -21,6 +37,11 @@ export const App: React.FC = () => {
             data-cy="perPageSelector"
             id="perPageSelector"
             className="form-control"
+            defaultValue={5}
```
VitaliyBodnar, line 40: move 5 to the constant as it will make a code cleaner

src/App.tsx
```diff
@@ -21,6 +37,11 @@ export const App: React.FC = () => {
...
+            onChange={(event) => {
+              setPerPage(+event.target.value);
+              setCurrentPage(1);
```
VitaliyBodnar, line 43: it is bad practise to make anonymous arrow functions in jsx.
make handler in component and pass reference here

check other places

src/App.tsx
```diff
@@ -34,62 +55,16 @@ export const App: React.FC = () => {
...
+        onChange={onPageChange}
+      />
+
       <ul>
-        <li data-cy="item">Item 1</li>
-        <li data-cy="item">Item 2</li>
-        <li data-cy="item">Item 3</li>
-        <li data-cy="item">Item 4</li>
-        <li data-cy="item">Item 5</li>
+        {items.slice(start, end).map(item => (
```
VitaliyBodnar, line 65: please don't call a function from the template because it will be executed every time when the state change, so make sense to move it to the variable

VitaliyBodnar (changes requested):

VitaliyBodnar (approved):

PR #657, opened 2023-10-06
Solution

andrii-rykal:
[DEMO LINK](https://andrii-rykal.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,26 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
-import { getNumbers } from './utils';
-
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
-  .map(n => `Item ${n}`);
+import { Pagination } from './components/Pagination';
 
 export const App: React.FC = () => {
+  const [currentPage, setCurrentPage] = useState(1);
+  const [perPage, setPerPage] = useState(5);
+
+  const TOTAL = 42;
+  const firstItem = 1 + perPage * (currentPage - 1);
+  const lastItem = Math.min((perPage + perPage * (currentPage - 1)), TOTAL);
+
+  const countPage = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setPerPage(+event.target.value);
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,18 +1,26 @@
...
-import { getNumbers } from './utils';
-
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
-  .map(n => `Item ${n}`);
+import { Pagination } from './components/Pagination';
 
 export const App: React.FC = () => {
+  const [currentPage, setCurrentPage] = useState(1);
+  const [perPage, setPerPage] = useState(5);
```
loralevitska, line 7: Try to avoid magic numbers (1 and 5). All numbers that are unclear from the context should be put into constants with a name that is understandable in terms of content

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,94 @@
...
+
+        <li className={cn('page-item', {
+          disabled: currentPage === amountPages,
+        })}
+        >
+          <a
+            data-cy="nextLink"
+            className="page-link"
+            href="#next"
+            aria-disabled={currentPage === amountPages ? 'true' : 'false'}
```
loralevitska, line 80:
```suggestion
            aria-disabled={currentPage === amountPages}
```

loralevitska (changes requested):
Almost perfect) Let's improve

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,98 @@
...
+          >
+            »
+          </a>
+        </li>
+      </ul>
+      <ul>
+        {amountItemsOnPage.map(item => (
+          <li data-cy="item" key={item}>{item}</li>
+        ))}
+      </ul>
```
etojeDenys, line 95: It should be outside of the pagination because this is not the pagination)

etojeDenys (approved):
good job

src/components/Pagination/Pagination.tsx
andrii-rykal: Please check if this is correct

PR #679, opened 2023-10-19
task implementation

yuliabakun:
[DEMO LINK](https://yulabakun.github.io/react_pagination/)

Some tests failed even though everything seems good to me on demo. 
Would appreciate some advice on how to fix those tests.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -37,14 +37,10 @@ export const App: React.FC = () => {
   const visibleItems = getVisibleItems(items, start, end);
   const titleInfo = `Page ${selectedPage} (items ${start + 1} - ${end + 1} of ${items.length})`;
 
-  const handleOnPageChange = (newPageNumber: number) => {
-    setSelectedPage(prevPage => {
-      if (prevPage !== newPageNumber) {
-        return newPageNumber;
-      }
-
-      return prevPage;
-    });
+  const handleOnPageChange = (
+    newPageNumber: number,
+  ) => {
+    setSelectedPage(newPageNumber);
   };
 
   return (
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,16 +4,26 @@ import { getNumbers } from './utils';
 import { Pagination } from './components/Pagination';
 
 type Item = {
-  id: number,
+  id: string,
   title: string,
 };
 
-const items = getNumbers(1, 42).map(n => `Item ${n}`);
+const items: Item[] = getNumbers(1, 42).map(n => {
+  const value = `Item ${n}`;
 
-const preparedItems: Item[] = items
-  .map((item, index) => ({ id: index, title: item }));
+  return {
+    id: value,
+    title: value,
+  };
+});
 
-function getVisibleItems(allItems: Item[], start: number, end: number) {
+const ITEMS_PER_PAGE = [3, 5, 10, 20];
+
+function getVisibleItems(
+  allItems: Item[],
+  start: number,
+  end: number,
+) {
   return allItems.slice(start, end + 1);
 }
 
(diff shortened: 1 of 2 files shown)


mpohorenyi (changes requested):
I'm sorry, but I'm rejecting the assignment because your solution failed 2 tests.
image

yulabakun:
👌 
Actually, I agreed with menthor to send it for review as it is and then they will help me to figure out what could cause tests failure because in demo everything works pretty much as expected. So I'm going to fix it then 😄
I will add this info to the pr description also 😉

src/App.tsx
```diff
@@ -1,18 +1,43 @@
...
-  .map(n => `Item ${n}`);
+type Item = {
+  id: number,
+  title: string,
+};
+
+const items = getNumbers(1, 42).map(n => `Item ${n}`);
+
+const preparedItems: Item[] = items
+  .map((item, index) => ({ id: index, title: item }));
```
polosanya, line 14: It is not a good idea to use index as id. Better generate some unique id using some tool like uuid or just use value as a key.
Also, there is no need to call `map` method twice, you could do it in the first one

src/App.tsx
```diff
@@ -1,18 +1,43 @@
...
+
+function getVisibleItems(allItems: Item[], start: number, end: number) {
```
polosanya, line 16:
```suggestion
function getVisibleItems(
  allItems: Item[], 
  start: number, 
  end: number,
) {
```

src/App.tsx
```diff
@@ -1,18 +1,43 @@
...
+  const [selectedPage, setSelectedPage] = useState(1);
+
+  const start = (selectedPage - 1) * itemsOnPage;
+  const end = Math.min(start + itemsOnPage - 1, preparedItems.length - 1);
+
+  const visibleItems = getVisibleItems(preparedItems, start, end);
+
+  const handleOnPageChange = (pageNumber: number) => {
+    if (selectedPage !== pageNumber) {
+      setSelectedPage(pageNumber);
```
polosanya, line 31: If you want to update the state according to the previous value, do it according to these rules:
https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state

src/App.tsx
```diff
@@ -1,18 +1,43 @@
...
+    }
+  };
+
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+        {`Page ${selectedPage} (items ${start + 1} - ${end + 1} of ${preparedItems.length})`}
```
polosanya, line 40: Would be great to create a variable for the text

src/App.tsx
```diff
@@ -21,6 +46,11 @@ export const App: React.FC = () => {
...
             id="perPageSelector"
             className="form-control"
+            defaultValue={itemsOnPage}
+            onChange={event => {
+              setitemsOnPage(+event.target.value);
+              handleOnPageChange(1);
+            }}
           >
             <option value="3">3</option>
             <option value="5">5</option>
```
polosanya, line 56: Not critical, but would be great to map through options instead of hardcoding

src/App.tsx
```diff
@@ -34,62 +64,17 @@ export const App: React.FC = () => {
...
+
       <ul>
-        <li data-cy="item">Item 1</li>
-        <li data-cy="item">Item 2</li>
-        <li data-cy="item">Item 3</li>
-        <li data-cy="item">Item 4</li>
-        <li data-cy="item">Item 5</li>
+        {visibleItems.map(item => (
+          <li data-cy="item" key={item.id}>{item.title}</li>
+        ))}
```
polosanya, line 77:
```suggestion
        {visibleItems.map(item => (
          <li data-cy="item" key={item.id}>
            {item.title}
          </li>
        ))}
```

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,69 @@
...
+  return (
+    <ul className="pagination">
+      <li className={cn('page-item',
+        { disabled: currentPage === pagesArray[0] })}
+      >
+        <a
+          data-cy="prevLink"
+          className="page-link"
+          href="#prev"
+          aria-disabled={currentPage === pagesArray[0] ? 'true' : 'false'}
```
polosanya, line 29: Create a variable for disabled condition so you can reuse it

src/App.tsx
```diff
@@ -1,18 +1,58 @@
...
+
+  const start = (selectedPage - 1) * itemsOnPage;
+  const end = Math.min(start + itemsOnPage - 1, items.length - 1);
+
+  const visibleItems = getVisibleItems(items, start, end);
+  const titleInfo = `Page ${selectedPage} (items ${start + 1} - ${end + 1} of ${items.length})`;
+
+  const handleOnPageChange = (newPageNumber: number) => {
+    setSelectedPage(prevPage => {
+      if (prevPage !== newPageNumber) {
```
StasSokolov, line 42: let's check if page is last, should help to pass the test

StasSokolov (changes requested):
Good job, let's try to fix failed test!

StasSokolov (approved):
Awesome job, i'm approving it🔥

PR #681, opened 2023-10-19
solution

OlegPopovych:
Hello. Check this sort of code, please, and I am going to implement router.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,62 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
-  .map(n => `Item ${n}`);
+
+const items: { title: string, id: number }[] = getNumbers(1, 42)
+  .map(n => {
+    return {
+      title: `Item ${n}`,
+      id: n,
+    };
+  });
+
+interface PagesOnPage<O> {
+  totalPages: number;
+  currentPage: O[];
+}
+
+const pagination = <T extends unknown>(
+  elements: T[],
+  itemsOnPage: number,
+  selectedPage: number,
+): PagesOnPage<T> => {
+  const totalPages = Math.ceil(elements.length / itemsOnPage);
+  const startIndex = 0 + (itemsOnPage * (selectedPage - 1));
+
+  const currentPage = [...elements].splice(startIndex, itemsOnPage);
+
+  const result: PagesOnPage<T> = {
+    totalPages,
+    currentPage,
+  };
+
+  return result;
+};
 
 export const App: React.FC = () => {
+  const [itemsOnPage, setItemsOnPage] = useState<number>(5);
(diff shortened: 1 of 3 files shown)


OlegPopovych:
The code version without Router passed tests successfully. With Router - not.

src/App.tsx
```diff
@@ -1,18 +1,102 @@
...
+import { useSearchParams } from 'react-router-dom';
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
-  .map(n => `Item ${n}`);
+
+const items: { title: string, id: number }[] = getNumbers(1, 42)
```
vasylmiskiv, line 9: You could create a separate interface for the object

src/App.tsx
```diff
@@ -1,18 +1,102 @@
...
+  Twenty = '20',
+}
+
+const pagination = <T extends unknown>(
+  elements: T[],
+  itemsOnPage: number,
+  selectedPage: number,
+): PagesOnPage<T> => {
+  const totalPages = Math.ceil(elements.length / itemsOnPage);
+  const startIndex = 0 + (itemsOnPage * (selectedPage - 1));
```
vasylmiskiv, line 35:
```suggestion
  const startIndex = itemsOnPage * (selectedPage - 1);
```

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,108 @@
...
+        <a
+          data-cy="pageLink"
+          className="page-link"
+          href={`#${n}`}
+          onClick={() => {
+            if (n !== currentPage) {
+              searchParams.set('page', String(n));
+              setSearchParams(searchParams);
+            }
+          }}
```
vasylmiskiv, line 37: If you want to use search params with pagination move the logic to  separate function instead of writing in the event

src/App.tsx
```diff
@@ -1,18 +1,102 @@
...
+      if (selectedPage < 0 || selectedPage > totalPages
+        || !Object.values(Object(Selects)).includes(String(itemsOnPage))) {
+        searchParams.set('page', '1');
+        searchParams.set('perPage', '5');
+        setSearchParams(searchParams);
+      }
+    }
+  }, [searchParams, setSearchParams]);
+
+  const firstElement = currentPage[0]?.id || 0;
```
vasylmiskiv, line 79: Why do you use optional chaining here? This is not necessary in this case

vasylmiskiv (approved):
Well done but look at the comments and don't forget to add link to your description

src/App.tsx
OlegPopovych: In this way, I set the default value in purpose to render Info even if a list is empty.

src/components/Pagination/Pagination.tsx
OlegPopovych: Done.

src/App.tsx
OlegPopovych: Done)

src/App.tsx
OlegPopovych: Done

OlegPopovych:
Tests were passed in PC

PR #693, opened 2023-10-28
solution

KseniyaGonchar:
https://KseniyaGonchar.github.io/react_pagination/

### DIFF
--- a/src/components/Pagination/Pagination.tsx
+++ b/src/components/Pagination/Pagination.tsx
@@ -1,1 +1,84 @@
-export const Pagination = () => {};
+/* eslint-disable @typescript-eslint/no-unused-vars */
+import React from 'react';
+import cn from 'classnames';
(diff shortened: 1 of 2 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,84 @@
...
+              href={`#${page}`}
+              onClick={() => handlePage(page)}
+            >
+              {page}
+            </a>
+          </li>
+        ))}
+        <li
+          className={cn('page-item', {
+            disabled: currentPage === lastPage,
```
volodymyr-soltys97, line 68: Consider will be better if you create a variable or helper function for this check and use it everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

SanyaBratashchuk (approved):

PR #694, opened 2023-10-29
Added pagination

IonShive:
# React Pagination

> Here is the [working version](https://ionshive.github.io/react_pagination/)


Esceype (approved):
Good job!

PR #699, opened 2023-10-30
add task solution

andrew17752985:
https://andrew17752985.github.io/react_pagination/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,37 @@
-import React from 'react';
+/* eslint-disable react/jsx-one-expression-per-line */
+/* eslint-disable max-len */
+import React, { useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
-
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
-  .map(n => `Item ${n}`);
+import { Pagination } from './components/Pagination';
 
 export const App: React.FC = () => {
+  const [currentPage, setCurrentPage] = useState<number>(1);
+  const [perPage, setPerPage] = useState<number>(5);
+
+  const totalItems = 42;
+  const items: string[] = getNumbers(1, totalItems).map((n) => `Item ${n}`);
+  const startIndex: number = (currentPage - 1) * perPage;
+  const endIndex: number = startIndex + perPage;
+  const displayedItems: string[] = items.slice(startIndex, endIndex);
+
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,18 +1,37 @@
...
 import { getNumbers } from './utils';
-
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
-  .map(n => `Item ${n}`);
+import { Pagination } from './components/Pagination';
 
 export const App: React.FC = () => {
+  const [currentPage, setCurrentPage] = useState<number>(1);
+  const [perPage, setPerPage] = useState<number>(5);
```
anastasiia-tilikina, line 10:
```suggestion
  const [currentPage, setCurrentPage] = useState(1);
  const [perPage, setPerPage] = useState(5);
```

you don't need to add type explicitly here as TS is already aware about type by your initial state

src/App.tsx
```diff
@@ -1,18 +1,37 @@
...
+
+  const totalItems = 42;
+  const items: string[] = getNumbers(1, totalItems).map((n) => `Item ${n}`);
+  const startIndex: number = (currentPage - 1) * perPage;
+  const endIndex: number = startIndex + perPage;
+  const displayedItems: string[] = items.slice(startIndex, endIndex);
```
anastasiia-tilikina, line 16:
```suggestion
  const items = getNumbers(1, totalItems).map((n) => `Item ${n}`);
  const startIndex = (currentPage - 1) * perPage;
  const endIndex = startIndex + perPage;
  const displayedItems = items.slice(startIndex, endIndex);
```

the same here

src/App.tsx
```diff
@@ -1,18 +1,37 @@
...
+
+  const handlePageChange = (newPage: number) => {
+    setCurrentPage(newPage);
+  };
+
+  const handlePerPageChange = (e: React.ChangeEvent<HTMLSelectElement>) => {
+    const newPerPage: number = parseInt(e.target.value, 10);
```
anastasiia-tilikina, line 23:
```suggestion
    const newPerPage = +e.target.value;
```
looks shorter, doesn't it?) (but your option is correct too)

anastasiia-tilikina (changes requested):
Well done!

let's jus fix some tiny issues)

maxim2310 (approved):
Good work!!!

PR #704, opened 2023-11-01
add task solution

mrogozhinskaya:
[DEMO LINK](https://mrogozhinskaya.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,35 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
+import { Pagination } from './components/Pagination/Pagination';
 import { getNumbers } from './utils';
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42)
   .map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const [currentPage, setCurrentPage] = useState(1);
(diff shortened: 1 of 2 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,96 @@
...
+          data-cy="prevLink"
+          className="page-link"
+          href="#prev"
+          aria-disabled={isPageFirst}
+          onClick={prevButtonClick}
+        >
+          «
+        </a>
+      </li>
+      {getButtons()}
```
etojeDenys, line 78: we can create the pages using the getNumbers function and just map it here. it will improve the code readability

src/App.tsx
```diff
@@ -21,75 +38,31 @@ export const App: React.FC = () => {
...
-          >
-            »
-          </a>
-        </li>
-      </ul>
+      <Pagination
+        total={items.length}
+        perPage={itemsPerPage}
+        currentPage={currentPage}
+        onPageChange={(page) => pageChanged(page)}
```
etojeDenys, line 59:
```suggestion
        onPageChange={pageChanged}
```

etojeDenys: actually we can use setCurrentPage directly

etojeDenys (changes requested):

etojeDenys (approved):
awesome

PR #738, opened 2023-11-28
Solution_1.0

andriy-fesych:
[DEMO LINK](https://andriy-fesych.github.io/react_pagination/)

You may change max items as well (added for fun 🙂), capped at 999.

### DIFF
--- a/src/components/Pagination/Pagination.tsx
+++ b/src/components/Pagination/Pagination.tsx
@@ -2,14 +2,11 @@ import classNames from 'classnames';
 import { useSearchParams } from 'react-router-dom';
 
 type Props = {
-  onPageChange: (value: number) => void,
-  current: number,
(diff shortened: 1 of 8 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,97 @@
...
+          </a>
+        </li>,
+      );
+    }
+
+    return pages;
+  }
+
+  return (
+    <>
```
etojeDenys, line 63: fragment looks redundant

etojeDenys (approved):
looks interesting, good job

src/components/Pagination/Pagination.tsx
andriy-fesych: true

PR #760, opened 2023-12-13
Develop

Patrykmclaren14:
[solution](https://Patrykmclaren14.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -11,6 +11,7 @@ export const App: React.FC = () => {
   const [perPage, setPerPage] = useState(5);
   const [currentPage, setCurrentPage] = useState(1);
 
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -21,6 +57,8 @@ export const App: React.FC = () => {
             data-cy="perPageSelector"
             id="perPageSelector"
             className="form-control"
+            value={perPage}
+            onChange={handlePerPage}
           >
             <option value="3">3</option>
```
mbruhler, line 63: Store these values in array somwhere above component and map them here

mbruhler (changes requested):
Overall good job, but:
- there is small change needed when you render options in select, please store them in some constant array and map options there.

JoyCoffeeAddict (approved):

PR #761, opened 2023-12-13
add task solution

Nataliia95254:
[DEMO LINK](https://Nataliia95254.github.io/react_pagination/) 

During testing faild 2 test. At the Q&A sessions mentor checked functionality and suggest do request becouse pagination works correct.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,59 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination/Pagination';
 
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -21,6 +62,12 @@ export const App: React.FC = () => {
             data-cy="perPageSelector"
             id="perPageSelector"
             className="form-control"
+            value={selectedValue}
+            onChange={(e: React.ChangeEvent<HTMLSelectElement>) => {
+              setActivePage(1);
+
+              return setSelectedValue(e.currentTarget.value);
+            }}
```
lerastarynets, line 70: Pls move the function inside `onChange` handler to a separate const and use as a callback here

PR #770, opened 2023-12-16
Create Pagination

VoloshynV:


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,104 @@
...
+    id: number,
+  ) => {
+    e.preventDefault();
+
+    onPageChange(id);
+  };
+
+  const handleNextPage = (
+    e: MouseEvent<HTMLAnchorElement, globalThis.MouseEvent>,
+
```
Viktor-Kost, line 29:
```suggestion
```

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,104 @@
...
+          data-cy="prevLink"
+          className="page-link"
+          href="#prev"
+          aria-disabled={isFirstPage}
+          onClick={handlePrevPage}
+        >
+          «
+        </a>
+      </li>
+      {new Array(totalPages).fill(null).map((_, idx) => {
```
Viktor-Kost, line 68: it's ok, yet you can make a little bit simplier
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from#specifications
https://stackoverflow.com/questions/3746725/how-to-create-an-array-containing-1-n

I like this variant
```
[...Array(totalPages).keys()]
```

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,104 @@
-export const Pagination = () => {};
+import { MouseEvent } from 'react';
+import cn from 'classnames';
+
+type Props = {
+  total: number;
+  perPage: number;
+  currentPage?: number;
```
Viktor-Kost, line 7: I see you're passing currentPage anyway, so no needs to make this prop optional

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,104 @@
...
+  onPageChange: (page: number) => void;
+};
+
+export const Pagination: React.FC<Props> = ({
+  total, perPage, currentPage = 1, onPageChange,
```
Viktor-Kost, line 12: so, an no needs to make default value. It's much better if you have this prop required
```suggestion
  total, perPage, currentPage, onPageChange,
```

Viktor-Kost (changes requested):
Great job!
Please provide DEMO link so we can check your result

src/components/Pagination/Pagination.tsx
VoloshynV: I have ts error for your variant: 
`Type 'IterableIterator<number>' can only be iterated through when using the '--downlevelIteration' flag or with a '--target' of 'es2015' or higher.ts(2802)`

But I use other variant from link.

src/components/Pagination/Pagination.tsx
VoloshynV: Yes, but this way the Pagination component will be less versatile for another case. 
Well, that was the requirement in the task)

Viktor-Kost: Agree. Haven't noticed this requirement
```
currentPage={1} /* optional with 1 by default */
```

Viktor-Kost (approved):
Well done! 🔥

PR #783, opened 2023-12-30
Develop

boytrend0108:
[DEMO LINK](https://boytrend0108.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,7 +4,7 @@ import { Pagination } from './components/Pagination';
 
 export const App: React.FC = () => {
   const [perPage, setPerPage] = useState('3');
-  const [currentPage, setcurrentPage] = useState('1');
+  const [currentPage, setCurrentPage] = useState(1);
 
   return (
     <div className="container">
@@ -38,7 +38,7 @@ export const App: React.FC = () => {
         total="42"
         perPage={perPage}
         currentPage={currentPage}
-        onPageChange={(page) => setcurrentPage(page)}
+        onPageChange={(page) => setCurrentPage(page)}
       />
     </div>
   );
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,26 +1,47 @@
...
+import { Pagination } from './components/Pagination';
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42)
-  .map(n => `Item ${n}`);
+  .map((n: number) => `Item ${n}`);
+
+function getCurrentItems(currentPage:number, perPage: number): string[] {
+  if (currentPage === 1) {
+    return [...items].splice(0, perPage);
```
TarasHoliuk, line 12: it's better to use `slice` in such cases - it doesn't mutate the original array, so no need to make a copy

src/App.tsx
```diff
@@ -34,62 +55,19 @@ export const App: React.FC = () => {
...
-          </a>
-        </li>
-      </ul>
+      <Pagination
+        total="42"
+        perPage={perPage}
+        currentPage={currentPage}
+        onPageChange={(page) => {
+          setCurrentPage(page);
+        }}
```
TarasHoliuk, line 64:
```suggestion
        onPageChange={setCurrentPage}
```
additional function wrapper is redundant

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,84 @@
...
+              if (currentPage !== 1) {
+                onPageChange(currentPage - 1);
+              }
+            }}
+          >
+            «
+          </a>
+        </li>
+
+        {[...totalPages].map(el => (
```
TarasHoliuk, line 47: no need to make a copy
`map` method doesn't change original array

TarasHoliuk (approved):
GJ :+1: 

Approved :heavy_check_mark: 
Added a few comments just FYI

PR #793, opened 2024-01-08
solution

InnaSh23:
[DEMO LINK](https://InnaSh23.github.io/react_pagination/)

### DIFF
--- a/src/components/Pagination/Pagination.tsx
+++ b/src/components/Pagination/Pagination.tsx
@@ -1,1 +1,96 @@
-export const Pagination = () => {};
+import React from 'react';
+import cn from 'classnames';
+import { getNumbers } from '../../utils';
+
+interface Props {
+  items: string[];
+  total: number;
+  perPage: number;
+  currentPage: number;
+  onPageChange: (value: number) => void;
+}
+
+export const Pagination: React.FC<Props> = ({
+  items,
+  total,
+  perPage,
+  currentPage = 1,
+  onPageChange,
+}) => {
+  const totalPages = Math.ceil(total / perPage);
+  const arrayItems: number[] = getNumbers(1, totalPages);
+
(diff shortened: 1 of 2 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,96 @@
...
+  perPage: number;
+  currentPage: number;
+  onPageChange: (value: number) => void;
+}
+
+export const Pagination: React.FC<Props> = ({
+  items,
+  total,
+  perPage,
+  currentPage = 1,
```
anastasiia-tilikina, line 17: current page is required prop not optional. so you don't need to set default value

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,96 @@
...
+      <ul className="pagination">
+        <li className={cn('page-item',
+          { disabled: currentPage === 1 })}
+        >
+          <a
+            data-cy="prevLink"
+            className="page-link"
+            href={currentPage !== 1 ? '#prev' : undefined}
+            aria-disabled={currentPage === 1}
+            onClick={currentPage !== 1
```
anastasiia-tilikina, line 40: I think no need to add a condition here since we've already checked in the function.  I'm not sure about passing tests though but try to avoid condition here

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,96 @@
...
+              : undefined}
+          >
+            »
+          </a>
+        </li>
+      </ul>
+
+      <ul>
+        {items
+          .slice((currentPage - 1) * perPage, currentPage * perPage)
```
anastasiia-tilikina, line 84: move this calculation to a var as well

anastasiia-tilikina (changes requested):
GJ!

just a few comments

PR #810, opened 2024-01-17
Task solution

tniklevych:
- [DEMO LINK](https://tniklevych.github.io/react_pagination/)

### DIFF
--- a/src/components/Pagination/Pagination.tsx
+++ b/src/components/Pagination/Pagination.tsx
@@ -1,1 +1,87 @@
-export const Pagination = () => {};
+import cn from 'classnames';
+import React, { useState } from 'react';
+
+interface PaginationProps {
+  total: number;
+  perPage: number;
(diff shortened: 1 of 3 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,87 @@
...
+
+  const handlePageChange = (pageNumber: number) => {
+    onPageChange(pageNumber, perPage);
+  };
+
+  const renderPageNumbers = () => {
+    const pageNumbers = [];
+
+    // eslint-disable-next-line no-plusplus
+    for (let i = 1; i <= totalPages; i++) {
```
IvanFesenko, line 27: const pages = new Array(totalPages).fill(null).map((_, index) => index + 1);

and than u may use this variable by map

IvanFesenko (approved):
well done

PR #835, opened 2024-02-05
solution

BodyaKutsyk:
[DEMO LINK](https://BodyaKutsyk.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,26 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
-import { getNumbers } from './utils';
+
+import { Pagination } from './components/Pagination';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
-  .map(n => `Item ${n}`);
-
 export const App: React.FC = () => {
+  const [perPage, setPerPage] = useState(5);
+  const [currentPage, setCurrentPage] = useState(1);
+  const total = 42;
+  const elemStart = perPage * (currentPage - 1) + 1;
+  let elemEnd = elemStart + perPage - 1;
+
+  if (elemEnd > total) {
+    elemEnd = total;
+  }
+
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+        {`Page ${currentPage} (items ${elemStart} - ${elemEnd} of ${total})`}
       </p>
 
       <div className="form-group row">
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,18 +1,26 @@
...
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
-  .map(n => `Item ${n}`);
+import { Pagination } from './components/Pagination';
 
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 export const App: React.FC = () => {
+  const [perPage, setPerPage] = useState(5);
+  const [currentPage, setCurrentPage] = useState(1);
```
witflash, line 9: It's better to move any "magic numbers" to the separate constant. 
For example:
```
const INITIAL_PER_PAGE = 5;
const INITIAL_PAGE = 1;
```

src/App.tsx
```diff
@@ -34,63 +47,13 @@ export const App: React.FC = () => {
...
-        <li data-cy="item">Item 2</li>
-        <li data-cy="item">Item 3</li>
-        <li data-cy="item">Item 4</li>
-        <li data-cy="item">Item 5</li>
-      </ul>
+      <Pagination
+        total={total}
+        perPage={perPage}
+        currentPage={currentPage}
+        onPageChange={(page) => setCurrentPage(page)}
```
witflash, line 54: If you use a handler that should be called with an argument it receives, you can use the shorter syntax:
```
onPageChange={setCurrentPage}
```
So, this way you pass the original function without creating an additional arrow function.

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,112 @@
...
+          <a
+            data-cy="prevLink"
+            className="page-link"
+            href="#prev"
+            aria-disabled={currentPage === 1}
+            onClick={() => {
+              if (currentPage > 1) {
+                onPageChange(currentPage - 1);
+              }
+            }}
```
witflash, line 43: It's better to move such handlers to the separate handlers:
```
const handlePageChangeClick = () => {
  if (currentPage > 1) {
    onPageChage(currentPage - 1);
  }
};

// and then just pass this function as a prop:

<a
  ...
  onClick={handlePageChangeClick}
/>
```

And your code will be cleaner and readable.

witflash: And you can use such an approach to any handlers below.

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,112 @@
...
+        {
+          itemsPerPage.map(n => {
+            if (n <= total) {
+              return (
+                <li data-cy="item">{`Item ${n}`}</li>
+              );
+            }
+
+            return false;
+          })
```
witflash, line 107: Do you really need this logic?
It looks like the `n` will never be larger than the `total` for this task.
But if you want to cover this corner case, it's better to do it the next way:
```
itemsPerPage
  .filter(value => value <= total)
  .map(n => (
    <li data-cy="item" key={n}>{`Item ${n}`}</li>
  ))
```

Pay attention that the element rendered by the `map` method should have the `key` prop.

witflash (changes requested):
Good job!

Pay attention to my comments to make your code clear.

SanyaBratashchuk (approved):

PR #836, opened 2024-02-06
solution

Dmytro-Kretov:
[DEMO LINK](https://dmytro-kretov.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,67 @@
-import React from 'react';
+import React, { useState } from 'react';
+
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
+import { ArrayNumber } from './Types/ArrayNumber';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42)
-  .map(n => `Item ${n}`);
+const items = getNumbers(1, 42);
+
+const options: ArrayNumber = [3, 5, 10, 20];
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,18 +1,67 @@
-import React from 'react';
+import React, { useState } from 'react';
+
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
+import { ArrayNumber } from './Types/ArrayNumber';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
lerastarynets, line 8: why did u disable this rule? consider removing unused vars

src/App.tsx
```diff
@@ -34,62 +89,20 @@ export const App: React.FC = () => {
...
-            aria-disabled="false"
-          >
-            »
-          </a>
-        </li>
-      </ul>
+      <Pagination
+        currentPage={currentPage}
+        pagesToShow={pagesToShow}
+        onPageChange={(page) => setCurrentPage(page)}
```
lerastarynets, line 95:
```suggestion
        onPageChange={setCurrentPage}
```

lerastarynets (approved):
Great job overall! Left a few comments to make your code more advanced)

src/App.tsx
Dmytro-Kretov: this rule was disabled from the beggining after installing the npm. so I didn't touch it. 

and now when I enable it it gives me error and i can't fix it

Dmytro-Kretov: image
and when i put ';' as expected - still it doesn't work. so that is why i left it as it was.

PR #849, opened 2024-02-15
task implemented

Kyryb:
[DEMO LINK](https://kyryb.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,7 +3,6 @@ import "./App.css";
 import { getNumbers } from "./utils";
 import { Pagination } from "./components/Pagination";
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items: string[] = getNumbers(1, 42).map((n) => `Item ${n}`);
 const totalItems: number = items.length;
 
@@ -16,6 +15,13 @@ export const App: React.FC = () => {
   if (lastItemNumber > totalItems) {
     lastItemNumber = totalItems;
   }
+
+  const handleItemsPerPageChange = (
+    event: React.ChangeEvent<HTMLSelectElement>,
+  ) => {
+    setPerPage(+event.target.value);
+    setCurrentPage(1);
+  };
 
   return (
     <div className="container">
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,18 +1,28 @@
-import React from 'react';
-import './App.css';
-import { getNumbers } from './utils';
+import React, { useState } from "react";
+import "./App.css";
+import { getNumbers } from "./utils";
+import { Pagination } from "./components/Pagination";
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
lerastarynets, line 6: why do u disable this rule? if var.const is unused, it should be deleted

src/App.tsx
```diff
@@ -21,6 +31,11 @@ export const App: React.FC = () => {
             data-cy="perPageSelector"
             id="perPageSelector"
             className="form-control"
+            value={perPage}
+            onChange={(e) => {
+              setPerPage(+e.target.value);
+              setCurrentPage(1);
+            }}
```
lerastarynets, line 38: Pls move the function inside `onChange` handler to a separate const and use as a callback here
**PLS DO THE SAME FOR ALL SUCH CASES IN THE PROJECT**

src/App.tsx
```diff
@@ -33,63 +48,16 @@ export const App: React.FC = () => {
...
-          >
-            »
-          </a>
-        </li>
-      </ul>
+      <Pagination
+        total={totalItems}
+        perPage={perPage}
+        currentPage={currentPage}
+        onPageChange={(page) => setCurrentPage(page)}
```
lerastarynets, line 55:
```suggestion
        onPageChange={setCurrentPage}
```

src/App.tsx
```diff
@@ -33,63 +48,16 @@ export const App: React.FC = () => {
...
+      />
       <ul>
-        <li data-cy="item">Item 1</li>
-        <li data-cy="item">Item 2</li>
-        <li data-cy="item">Item 3</li>
-        <li data-cy="item">Item 4</li>
-        <li data-cy="item">Item 5</li>
+        {getNumbers(firstItemNumber, lastItemNumber).map((item) => {
+          return <li data-cy="item">Item {item}</li>;
+        })}
```
lerastarynets, line 60:
```suggestion
        {getNumbers(firstItemNumber, lastItemNumber).map((item) => (
         <li data-cy="item">Item {item}</li>;
        ))}
```
also where is `key` prop?

**PLS DO THE SAME FOR ALL SUCH CASES IN THE PROJECT**

lerastarynets (changes requested):
Nice start! Let's fix the solution according to the checklist

etojeDenys (approved):
lgtm

PR #862, opened 2024-03-01
solution

Yehorf21:
[DEMO LINK](https://Yehorf21.github.io/react_pagination/)

I handled components a bit differently than the instructions said, but the output is the same


l4st1m0za (approved):
Good job!

PR #867, opened 2024-03-03
add task solution

creonluv:
[DEMO LINK](https://creonluv.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,34 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42).map(n => `Item ${n}`);
 
+enum PerPageAmount {
+  Three = 3,
+  Five = 5,
+  Ten = 10,
+  Twenty = 20,
+}
+
 export const App: React.FC = () => {
+  const [currentPage, setCurrentPage] = useState(1);
+  const [perPage, setPerPage] = useState(PerPageAmount.Five);
+
+  const totalAmount = items.length;
+  const firstIndex = (currentPage - 1) * perPage;
+  const lastIndex = Math.min(currentPage * perPage, totalAmount);
+  const currentItems = items.slice(firstIndex, lastIndex);
+
   return (
     <div className="container">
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,29 +1,52 @@
...
+            value={perPage}
+            onChange={event => {
+              setCurrentPage(1);
+              setPerPage(+event.target.value);
+            }}
+          >
+            <option value="3">{PerPageAmount.Three}</option>
+            <option value="5">{PerPageAmount.Five}</option>
+            <option value="10">{PerPageAmount.Ten}</option>
+            <option value="20">{PerPageAmount.Twenty}</option>
```
VitaliyBondarenko1982, line 49:
```suggestion
            {[3, 5, 10, 20].map(option => <option key={option} value={option}>{option}</option>)}
```
looks like redundant to make enum for options.
suggest to make array and map all options.
something like that

VitaliyBondarenko1982 (changes requested):
Looks good.
Suggest a little refactoring

src/App.tsx
```diff
@@ -1,29 +1,46 @@
...
-            <option value="3">3</option>
-            <option value="5">5</option>
-            <option value="10">10</option>
-            <option value="20">20</option>
+            className="form-control"
+            value={perPage}
+            onChange={event => {
+              setCurrentPage(1);
+              setPerPage(+event.target.value);
+            }}
```
Viktor-Kost, line 37: it's better to move this `onChange` handler to a variable

src/App.tsx
```diff
@@ -1,29 +1,46 @@
...
+          >
+            {[3, 5, 10, 20].map(option => (
```
Viktor-Kost, line 39: move an array to a variable

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,74 @@
...
+
+  const nextPage = () => {
+    if (currentPage < pagesAmount) {
+      onChangePage(currentPage + 1);
+    }
+  };
+
+  return (
+    <ul className="pagination">
+      <li className={`page-item ${currentPage === 1 ? 'disabled' : ''}`}>
```
Viktor-Kost, line 31: it' better to use a library for conditional class names
Example:
https://www.npmjs.com/package/classnames

Viktor-Kost (approved):
GJ!
just apply recommendations for yourself

PR #878, opened 2024-03-09
Solution

wsadocha:
[DEMO LINK](https://wsadocha.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,32 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42).map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const [perPage, setPerPage] = useState('5');
+  const [currentPage, setCurrentPage] = useState(1);
+  const firstItem = +perPage * +currentPage - +perPage + 1;
+  let lastItem = +perPage * +currentPage;
+
+  const handlePerPageChange = (e: React.ChangeEvent<HTMLSelectElement>) => {
+    setCurrentPage(1);
+    setPerPage(e.target.value);
+  };
+
+  if (lastItem > items.length) {
+    lastItem = items.length;
+  }
+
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,25 +1,43 @@
...
+import React, { useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42).map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const [perPage, setPerPage] = useState('5');
```
yevhenii-pyl, line 10: Why bothering with strings and converting string => number? 
You are the boss here, make these values numbers by the default.

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,98 @@
...
+  perPage,
+  currentPage,
+  onPageChange,
+}: Props) => {
+  const amountOfPages = Math.ceil(total / +perPage);
+  const pageList = [];
+  const itemList = [];
+
+  // eslint-disable-next-line no-plusplus
+  for (let i = 1; i <= amountOfPages; i++) {
```
yevhenii-pyl, line 21: That looks like an overcomplication here, why don't you just create pages: 
`  const pages = new Array<number>(pagesAmount).fill(0).map((_, i) => i + 1);`

and then map through them creating the same li > a structure inside your render?

yevhenii-pyl (changes requested):
GJ 👍 
But let's refactor a bit to make it even cleaner

src/App.tsx
```diff
@@ -1,29 +1,61 @@
...
-            className="form-control">
-            <option value="3">3</option>
-            <option value="5">5</option>
-            <option value="10">10</option>
-            <option value="20">20</option>
+            className="form-control"
+            value={perPage}
+            onChange={handlePerPageChange}
+          >
+            {Object.values(PerPageAmount).map(option => {
```
yevhenii-pyl, line 48: Interesting approach :) 
A bit of overcomplication here I guess

yevhenii-pyl (approved):
Yep, you got the idea. Nice job 🥇

PR #900, opened 2024-03-17
Solution

ivan-gritsenko:
[DEMO LINK](https://ivan-gritsenko.github.io/react_pagination/)


olduma (approved):
Generally good, but some tests did not pass, because of missing key of mapping.

Check failure on line 61 in src/components/Pagination/Pagination.tsx
Missing "key" prop for element in iterator

BudnikOleksii (approved):
Good work 👍

PR #975, opened 2024-05-09
task solution

MatheusStefan:
[DEMO LINK](https://MatheusStefan.github.io/react_pagination/)


EdPirro (changes requested):
Great job, @MatheusStefan!

Just a thing to keep in mind: when allowing search params any value is possible. Remember to validate the initial state as well!

For instance try visiting https://matheusstefan.github.io/react_pagination/?page=100&perPage=10#

Even negative values would be possible: 
- https://matheusstefan.github.io/react_pagination/?page=-1&perPage=10
- https://matheusstefan.github.io/react_pagination/?page=1&perPage=-1

Regardless, congratulations on taking up this challenge 👏

PR #983, opened 2024-05-17
add task solution

sophiasph:
[DEMO LINK](https://sophiasph.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,29 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,29 +1,54 @@
...
           <select
             data-cy="perPageSelector"
             id="perPageSelector"
-            className="form-control">
-            <option value="3">3</option>
-            <option value="5">5</option>
-            <option value="10">10</option>
-            <option value="20">20</option>
+            className="form-control"
+            onChange={event => {
```
lerastarynets, line 35: Pls move the function inside `onChange` handler to a separate const and use as a callback here
*PLS DO THE SAME FOR ALL SUCH CASES IN THE PROJECT**

lerastarynets (approved):
Great job overall! Left one tiny suggestion for the improvement

PR #984, opened 2024-05-18
solution

YuliiaCat:
https://yuliiacat.github.io/react_pagination/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,28 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42).map(n => `Item ${n}`);
+const total = items.length;
 
 export const App: React.FC = () => {
+  const [itemsPerPage, setItemsPerPage] = useState(5);
+  const [currPage, setCurrPage] = useState(1);
+  const numOfPages = Math.ceil(items.length / itemsPerPage);
+
+  const elements = currPage * itemsPerPage;
+
+  const fromItem = (currPage - 1) * itemsPerPage + 1;
+  const toItem = elements > total ? total : elements;
+
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,29 +1,53 @@
...
+  const fromItem = (currPage - 1) * itemsPerPage + 1;
+  const toItem = elements > total ? total : elements;
+
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+        Page {currPage} (items {fromItem} - {toItem} of {total})
```
volodymyr-soltys97, line 25: You can move this logic to the helper variable and use it here

src/App.tsx
```diff
@@ -1,29 +1,53 @@
...
+            onChange={event => {
+              setItemsPerPage(+event.target.value);
+              setCurrPage(() => {
+                const defaultPage = 1;
+
+                window.location.hash = `#${defaultPage}`;
+
+                return defaultPage;
+              });
+            }}
```
volodymyr-soltys97, line 43: Move this logic from jsx to the helper function and use it here

src/App.tsx
```diff
@@ -1,29 +1,53 @@
...
+            defaultValue={5}
+          >
+            {[3, 5, 10, 20].map((option, numb) => (
+              <option value={option} key={numb}>
```
volodymyr-soltys97, line 47: Move this array to the variable and use it here
```suggestion
            {[3, 5, 10, 20].map((option) => (
              <option value={option} key={option}>
```

src/App.tsx
```diff
@@ -32,78 +56,27 @@ export const App: React.FC = () => {
...
+        currentPage={currPage}
+        onPageChange={page => {
+          if (page === currPage) {
+            return;
+          }
+
+          if (page >= 1 && page <= numOfPages) {
+            setCurrPage(page);
+          }
+        }}
```
volodymyr-soltys97, line 71: Move this logic from jsx to the helper function and use it here

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

vadiimvooo (approved):
Well done!)

PR #993, opened 2024-05-28
solution

yanamomot:
[DEMO LINK](https://yanamomot.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,50 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
+import { Pagination } from './components/Pagination';
+import { PerPage } from './PerPage';
 import { getNumbers } from './utils';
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,29 +1,65 @@
...
-            className="form-control">
-            <option value="3">3</option>
-            <option value="5">5</option>
-            <option value="10">10</option>
-            <option value="20">20</option>
+            className="form-control"
+            value={perPage}
+            onChange={event => handleSelectCountItems(event.target.value)}
+          >
+            <option value={PerPage.three}>3</option>
```
mrkovaliv, line 59: use map for this goal

`const perPages = [PerPage.three, PerPage.five ...];`

and then `perPages.map(perPage =>  <option value={perPage}>perPage</option>)`

PR #1001, opened 2024-06-11
Develop

arthurfujii:
[DEMO LINK](https://arthurfujii.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -20,6 +20,7 @@ export const App: React.FC = () => {
 
   const handlePerPage = (event: React.ChangeEvent<HTMLSelectElement>) => {
     setPerPage(+event.target.value);
+    setCurrentPage(1);
   };
 
   return (
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,33 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42).map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,25 +1,43 @@
...
+  const numberPages: number[] = getPageNumbers(items.length, perPage);
+  const currentIndex = numberPages.indexOf(currentPage);
+  const fromIndex = currentIndex * perPage;
+  const toIndex =
+    fromIndex + perPage <= items.length ? fromIndex + perPage : items.length;
+
+  const handlePerPage = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setPerPage(+event.target.value);
+    setCurrentPage(1);
+  };
```
alysonvilela, line 22: Congratulations!

Nice logic you got!
A next step here is copy all these code and make a custom hook `usePagination` that exports the used constants. (you don't need to do it now, just study the concept)

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,77 @@
...
+  onPageChange: (page: number) => void;
+};
+
+export const Pagination: React.FC<PaginationProps> = ({
+  total,
+  perPage,
+  currentPage,
+  onPageChange,
+}) => {
+  const totalPages: number[] = getPageNumbers(total, perPage);
```
alysonvilela, line 17: You could get the result of this function on the higher component, and here just receive totalPages prop

alysonvilela (approved):

src/App.tsx
arthurfujii: Thanks! 
Custom hook is new to me! I'll take a look how it works!! Thank you!

src/components/Pagination/Pagination.tsx
arthurfujii: I've thought in this solution, but at ReadMe file, they ask the specifics props, so I used this way.

PR #1005, opened 2024-06-17
add task solution

kravets111:
[DEMO LINK](https://kravets111.github.io/react_pagination/)


lerastarynets (approved):
You're on the right track!

PR #1020, opened 2024-07-29
add solution

smachylooo:
[DEMO LINK](https://smachylooo.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,42 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
+import { Pagination } from './components/Pagination';
 import { getNumbers } from './utils';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42).map(n => `Item ${n}`);
+const items = getNumbers(1, 42).map(el => `Item ${el}`);
 
 export const App: React.FC = () => {
+  const [currentPage, setCurrentPage] = useState(1);
+  const [currentOption, setCurrentOption] = useState(5);
+
+  function handleOpptionValue(e: React.ChangeEvent<HTMLSelectElement>) {
+    setCurrentPage(1);
+    setCurrentOption(+e.target.value);
+  }
+
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+        Page {currentPage} (items{' '}
+        {currentPage * currentOption - currentOption + 1} -{' '}
+        {items.length > currentPage * currentOption
+          ? currentPage * currentOption
+          : items.length}{' '}
+        of {items.length})
       </p>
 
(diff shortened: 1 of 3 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,84 @@
...
+          <a
+            data-cy="nextLink"
+            className="page-link"
+            href="#next"
+            onClick={e => {
+              e.preventDefault();
+              if (currentPage < pagination && onPageChange) {
+                onPageChange(currentPage + 1);
+              }
+            }}
```
lerastarynets, line 64: Pls move the function inside `onClick` handler to a separate const and use as a callback here
**PLS DO THE SAME FOR ALL SUCH CASES IN THE PROJECT**

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,84 @@
...
+          >
+            »
+          </a>
+        </li>
+      </ul>
+      <ul>
+        {Array.from({ length: perPage }, (_, index) => {
+          const item = total[currentPage * perPage - perPage + index];
+
+          return item !== undefined ? (
```
lerastarynets, line 75:
```suggestion
          return !!item ? (
```

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,92 @@
...
+
+  return (
+    <>
+      <ul className="pagination">
+        <li className={`page-item ${currentPage === 1 ? 'disabled' : ''}`}>
+          <a
+            data-cy="prevLink"
+            className="page-link"
+            href="#prev"
+            onClick={e => onClickHandler(e, 1)}
```
etojeDenys, line 44: better to create separate function for prev and next, instead of one complex with strange checks

src/App.tsx
```diff
@@ -1,25 +1,42 @@
...
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+        Page {currentPage} (items{' '}
+        {currentPage * currentOption - currentOption + 1} -{' '}
+        {items.length > currentPage * currentOption
+          ? currentPage * currentOption
```
etojeDenys, line 26: always better to move all calculation before the jsx

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,92 @@
...
+            aria-disabled={currentPage === 1 ? 'true' : 'false'}
+          >
+            «
+          </a>
+        </li>
+        {Array.from({ length: pagination }, (_, index) => {
```
etojeDenys, line 50: you can reuse the getNumbers function here

etojeDenys (changes requested):
few comments

PR #1043, opened 2024-09-02
add task solution

siefimov:
[DEMO LINK](https://siefimov.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,50 @@
-import React from 'react';
+import { useEffect, useState } from 'react';
+import { useLocation, useNavigate } from 'react-router-dom';
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42).map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const location = useLocation();
+  const navigate = useNavigate();
+
+  const queryParams = new URLSearchParams(location.search);
+  const initialPage = parseInt(queryParams.get('page') || '1', 10);
+  const initialPerPage = parseInt(queryParams.get('perPage') || '5', 10);
+
+  const [currentPage, setCurrentPage] = useState(initialPage);
+  const [perPage, setPerPage] = useState(initialPerPage);
+
+  useEffect(() => {
+    const params = new URLSearchParams();
+
+    params.set('page', currentPage.toString());
+    params.set('perPage', perPage.toString());
+    navigate({ search: `?${params.toString()}` });
+  }, [currentPage, perPage, navigate]);
+
+  const startIndex = (currentPage - 1) * perPage;
+  const currentItems = items.slice(startIndex, startIndex + perPage);
+
+  const handlePageChange = (page: number) => {
+    setCurrentPage(page);
+  };
+
+  const handlePerPageChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setPerPage(parseInt(event.target.value, 10));
+    setCurrentPage(1);
+  };
+
   return (
     <div className="container">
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -32,78 +68,19 @@ export const App: React.FC = () => {
...
+      />
+
       <ul>
-        <li data-cy="item">Item 1</li>
-        <li data-cy="item">Item 2</li>
-        <li data-cy="item">Item 3</li>
-        <li data-cy="item">Item 4</li>
-        <li data-cy="item">Item 5</li>
+        {currentItems.map((item, index) => (
+          <li key={index} data-cy="item">
```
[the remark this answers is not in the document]
TarasHoliuk: In this case it's almost the same, but anyway, it's better to use item as a key instead of index

src/App.tsx
```diff
@@ -1,25 +1,62 @@
...
+
+    params.set('page', currentPage.toString());
+    params.set('perPage', perPage.toString());
+    navigate({ search: `?${params.toString()}` });
+  }, [currentPage, perPage, navigate]);
+
+  const startIndex = (currentPage - 1) * perPage;
+  const currentItems = items.slice(startIndex, startIndex + perPage);
+
+  const handlePageChange = (page: number) => {
```
TarasHoliuk, line 33: redundant function wrapper

just use `setCurrentPage`

src/App.tsx
```diff
@@ -1,25 +1,62 @@
...
+  const [perPage, setPerPage] = useState(initialPerPage);
+
+  const items = useMemo(() => getNumbers(1, 42).map(n => `Item ${n}`), []);
+
+  useEffect(() => {
+    const params = new URLSearchParams();
+
+    params.set('page', currentPage.toString());
+    params.set('perPage', perPage.toString());
+    navigate({ search: `?${params.toString()}` });
```
TarasHoliuk, line 27: just use `a` tag with the correct href as it was initially
So this entire useEffect is redundant

src/App.tsx
```diff
@@ -1,25 +1,62 @@
...
+  };
+
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+        Page {currentPage} (items {startIndex + 1} -{' '}
+        {Math.min(startIndex + perPage, items.length)} of {items.length})
```
TarasHoliuk, line 48: save result in a variable above and just use here instead of doing this computation is JSX

TarasHoliuk (changes requested):

TarasHoliuk:
Try to fix tests

src/App.tsx
```diff
@@ -1,25 +1,50 @@
...
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42).map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const location = useLocation();
+
+  const queryParams = new URLSearchParams(location.search);
+  const initialPage = parseInt(queryParams.get('page') || '1', 10);
+  const initialPerPage = parseInt(queryParams.get('perPage') || '5', 10);
```
etojeDenys, line 14: you can simplify it just using the useSearchParams hook

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,74 @@
...
+  const pageNumbers = getNumbers(1, totalPages);
+
+  const handlePageChange = (page: number) => {
+    if (page !== currentPage && page > 0 && page <= totalPages) {
+      onPageChange(page);
+    }
+  };
+
+  return (
+    <>
```
etojeDenys, line 28: fragment is redundant here 

https://react.dev/reference/react/Fragment#usage

etojeDenys (approved):

PR #1058, opened 2024-09-25
solution

Daryna-Kukharets:
[DEMO LINK](https://Daryna-Kukharets.github.io/react_pagination/)


etojeDenys (approved):
looks different than in the working example
image


but the logic looks good

PR #1068, opened 2024-09-30
solution

YuliiaShatkovska:
[DEMO LINK](https://YuliiaShatkovska.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,38 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42).map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const [selectedPerPageValue, setSelectedPerPageValue] = useState<number>(5);
+  const [currentPage, setCurrentPage] = useState<number>(1);
+
+  const handlePerPageChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setSelectedPerPageValue(Number(event.target.value));
+    setCurrentPage(1);
+  };
+
+  const onPageChange = (page: number) => {
+    setCurrentPage(page);
+  };
+
+  const indexOfLastItem = currentPage * selectedPerPageValue;
+  const indexOfFirstItem = indexOfLastItem - selectedPerPageValue;
+
+  const currentItems = items.slice(indexOfFirstItem, indexOfLastItem);
+
+  const itemFrom = indexOfFirstItem + 1;
+  const itemTo = Math.min(indexOfLastItem, items.length);
+
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+        Page {currentPage} (items {itemFrom} - {itemTo} of {items.length})
       </p>
 
       <div className="form-group row">
@@ -19,7 +40,10 @@ export const App: React.FC = () => {
           <select
             data-cy="perPageSelector"
             id="perPageSelector"
-            className="form-control">
+            className="form-control"
+            onChange={handlePerPageChange}
+            defaultValue={selectedPerPageValue}
+          >
             <option value="3">3</option>
             <option value="5">5</option>
             <option value="10">10</option>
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,25 +1,49 @@
...
+  const [currentPage, setCurrentPage] = useState<number>(1);
+
+  const handlePerPageChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setSelectedPerPageValue(Number(event.target.value));
+    setCurrentPage(1);
+  };
+
+  const onPageChange = (page: number) => {
+    setCurrentPage(page);
+  };
```
hma-3: `onPageChange` callback should be triggered only if page was changed

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,82 @@
...
+  onPageChange: (value: number) => void;
+};
+
+export const Pagination: FC<Props> = ({
+  total,
+  perPage,
+  currentPage,
+  onPageChange,
+}) => {
+  const amountOfPages = getNumbers(1, Math.ceil(total / perPage));
```
hma-3, line 18: Maybe you should rename this variable, because it's an array of all pages, not their amount

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,82 @@
...
+    }
+
+    if (direction === 'prev' && currentPage !== 1) {
+      onPageChange(currentPage - 1);
+    }
+  };
+
+  return (
+    <ul className="pagination">
+      <li className={cn('page-item', { disabled: currentPage === 1 })}>
```
hma-3: `currentPage === 1` -- you have already used it several times, consider creating a variable for it
same for `currentPage === amountOfPages.length`

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,82 @@
...
+  const handlePageChange = (
+    event: React.MouseEvent<HTMLLIElement, MouseEvent>,
+  ) => {
+    const element = event.target as HTMLElement;
+    const page = element.textContent;
+
+    onPageChange(Number(page));
+  };
+
+  const handleArrowPageChange = (direction: string) => {
```
hma-3: `direction: string` -- there can be only 'prev' or 'next', so you can use union type: `direction: 'prev' | 'next'`

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,82 @@
...
+        >
+          «
+        </a>
+      </li>
+
+      {amountOfPages.map(page => (
+        <li
+          className={cn('page-item', { active: currentPage === page })}
+          key={page}
+          onClick={handlePageChange}
```
hma-3: You always write 'onClick' event inside `<a>`, so why here it's inside `<li>`?

hma-3 (approved):
Grate job!
Have just a few suggestions below ⬇️

src/App.tsx
```diff
@@ -1,25 +1,49 @@
...
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const items = getNumbers(1, 42).map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const [selectedPerPageValue, setSelectedPerPageValue] = useState<number>(5);
+  const [currentPage, setCurrentPage] = useState<number>(1);
```
VolodymyrKirichenko: if u set the default value to a simple type - typescript itself understands what type it is, so it is unnecessary.
For all such cases
```suggestion
  const [selectedPerPageValue, setSelectedPerPageValue] = useState(5);
  const [currentPage, setCurrentPage] = useState(1);
```

src/App.tsx
VolodymyrKirichenko: Good point 🔥 
U need additional condition for it.
Do it

src/App.tsx
```diff
@@ -1,25 +1,49 @@
...
+  };
+
+  const onPageChange = (page: number) => {
+    setCurrentPage(page);
+  };
+
+  const indexOfLastItem = currentPage * selectedPerPageValue;
+  const indexOfFirstItem = indexOfLastItem - selectedPerPageValue;
+
+  const currentItems = items.slice(indexOfFirstItem, indexOfLastItem);
```
VolodymyrKirichenko: Use the function - getNumbers immediately.
```suggestion
  const currentItems = getNumbers(indexOfFirstItem, indexOfLastItem)).map(n => `Item ${n}`;
```

src/App.tsx
```diff
@@ -1,25 +1,49 @@
...
             data-cy="perPageSelector"
             id="perPageSelector"
-            className="form-control">
+            className="form-control"
+            onChange={handlePerPageChange}
+            defaultValue={selectedPerPageValue}
+          >
             <option value="3">3</option>
             <option value="5">5</option>
             <option value="10">10</option>
```
VolodymyrKirichenko: map?

src/App.tsx
```diff
@@ -1,25 +1,49 @@
...
 
       <div className="form-group row">
         <div className="col-3 col-sm-2 col-xl-1">
           <select
             data-cy="perPageSelector"
             id="perPageSelector"
-            className="form-control">
+            className="form-control"
+            onChange={handlePerPageChange}
+            defaultValue={selectedPerPageValue}
```
VolodymyrKirichenko: use `value` instead `defaultValue`

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,82 @@
...
+export const Pagination: FC<Props> = ({
+  total,
+  perPage,
+  currentPage,
+  onPageChange,
+}) => {
+  const amountOfPages = getNumbers(1, Math.ceil(total / perPage));
+
+  const handlePageChange = (
+    event: React.MouseEvent<HTMLLIElement, MouseEvent>,
```
VolodymyrKirichenko:
```suggestion
    event: React.MouseEvent<HTMLElement>,
```

src/components/Pagination/Pagination.tsx
VolodymyrKirichenko: Good point 🔥 
Use `union type` in this case

src/components/Pagination/Pagination.tsx
VolodymyrKirichenko: Good point 🚀 
create variable

VolodymyrKirichenko (changes requested):
Good job 👍 
But let's make some changes 🥹

src/components/Pagination/Pagination.tsx
igorkruz: not fixed!

igorkruz (approved):
Well done 🔥

PR #1074, opened 2024-10-02
Solution

KrivorotkoYurii:
[DEMO LINK](https://KrivorotkoYurii.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,38 @@
-import React from 'react';
+import React, { useState } from 'react';
+
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42).map(n => `Item ${n}`);
+const options = [3, 5, 10, 20];
 
 export const App: React.FC = () => {
+  const [perPageCount, setPerPageCount] = useState(5);
+  const [currentPage, setCurrentPage] = useState(1);
+
+  const totalItems = 42;
+
+  const firstItemOnPage = (currentPage - 1) * perPageCount;
+
+  const lastItemOnPage = firstItemOnPage + perPageCount;
+
+  const amountOfItems = getNumbers(1, 42).filter(
+    a => a >= firstItemOnPage + 1 && a <= lastItemOnPage,
+  );
+
+  const visibleItems = amountOfItems.map(n => `Item ${n}`);
+
+  const handleChangeOption = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setPerPageCount(+event.target.value);
+    setCurrentPage(1);
+  };
+
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+        {`Page ${currentPage} (items ${amountOfItems[0]} - ${amountOfItems[amountOfItems.length - 1]} of 42)`}
       </p>
 
       <div className="form-group row">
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,29 +1,54 @@
...
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42).map(n => `Item ${n}`);
+const options = [3, 5, 10, 20];
 
 export const App: React.FC = () => {
+  const [perPageCount, setPerPageCount] = useState(5);
+  const [currentPage, setCurrentPage] = useState(1);
+
+  const totalItems = 42;
```
VolodymyrKirichenko: u can use CONSTANTS for it.
in ur task - it will never change

src/App.tsx
```diff
@@ -1,29 +1,54 @@
...
+
+  const firstItemOnPage = (currentPage - 1) * perPageCount;
+
+  const lastItemOnPage = firstItemOnPage + perPageCount;
```
VolodymyrKirichenko: first/last item on page, or first/last item index?

src/App.tsx
```diff
@@ -1,29 +1,54 @@
...
+
+  const amountOfItems = getNumbers(1, 42).filter(
+    a => a >= firstItemOnPage + 1 && a <= lastItemOnPage,
```
VolodymyrKirichenko: `a`?What the `a`?
`anaconda`? 😶‍🌫️ 
change it

src/App.tsx
```diff
@@ -1,29 +1,54 @@
...
+  const [perPageCount, setPerPageCount] = useState(5);
+  const [currentPage, setCurrentPage] = useState(1);
+
+  const totalItems = 42;
+
+  const firstItemOnPage = (currentPage - 1) * perPageCount;
+
+  const lastItemOnPage = firstItemOnPage + perPageCount;
+
+  const amountOfItems = getNumbers(1, 42).filter(
```
VolodymyrKirichenko: u can get only the items u need on a current page. u don't get everything at once

src/App.tsx
```diff
@@ -1,29 +1,54 @@
...
+    a => a >= firstItemOnPage + 1 && a <= lastItemOnPage,
+  );
+
+  const visibleItems = amountOfItems.map(n => `Item ${n}`);
```
VolodymyrKirichenko: `n`? Change for all such cases

src/App.tsx
```diff
@@ -1,29 +1,54 @@
...
             data-cy="perPageSelector"
             id="perPageSelector"
-            className="form-control">
-            <option value="3">3</option>
-            <option value="5">5</option>
-            <option value="10">10</option>
-            <option value="20">20</option>
+            className="form-control"
+            value={perPageCount}
+            onChange={event => handleChangeOption(event)}
```
VolodymyrKirichenko: `event `- is sent by default, you don't need to specify it
```suggestion
            onChange={handleChangeOption}
```

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,80 @@
...
+  };
+
+  return (
+    <ul className="pagination">
+      <li className={cn('page-item', { disabled: currentPage === 1 })}>
+        <a
+          data-cy="prevLink"
+          className="page-link"
+          href="#prev"
+          aria-disabled={currentPage === 1 ? 'true' : 'false'}
```
VolodymyrKirichenko: comparison always returns a boolean value. u don't need to add a ternary operator here. It will be the same. 
Fix it for all such cases
```suggestion
          aria-disabled={currentPage === 1}
```

VolodymyrKirichenko (changes requested):
I think this is ur old commit, because we fixed it on **Q&A** a day ago
In any case, it must be corrected

BudnikOleksii (approved):
Looks good 🚀

PR #1075, opened 2024-10-03
develop

Ostkreuzzz:
https://Ostkreuzzz.github.io/react_pagination/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,29 +1,64 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
+import cn from 'classnames';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,29 +1,64 @@
...
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+        {`Page ${currentPage} (items ${firstIndex + 1} - ${lastItemIndex} of ${items.length})`}
       </p>
 
       <div className="form-group row">
         <div className="col-3 col-sm-2 col-xl-1">
           <select
+            defaultValue="5"
```
v1RnT, line 51: As i know, it's better not to use `defaultValue` when you're also managing its state. It'd be better to use `value` with element from state.

Something like this:
```suggestion
            value={perPage}
```

v1RnT (approved):
Looks good, just check the comment I left.

PR #1078, opened 2024-10-13
Solved

NazariiAlieksieiev:
- [DEMO](https://nazariialieksieiev.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,32 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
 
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,25 +1,44 @@
...
+    setCurrentPage(1);
+  };
+
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+        Page {currentPage} (items {start + 1} - {end} of {totalNumber})
```
volodymyr-soltys97, line 30: You can move this logic from jsx to the helper variable and use it here

volodymyr-soltys97 (approved):
Great work! 🔥

PR #1091, opened 2024-10-31
Solution

ab3MN:
[DEMO LINK](https://ab3MN.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -5,7 +5,6 @@ import { getItemsOnPage } from './utils/pagination/getItemsOnPage';
 import { getItemsOnPageCount } from './utils/pagination/getItemsOnPageCount';
 import { Pagination } from './components/Pagination';
 import { EPaginationSize } from './enums/Pagination';
-import { getTotalPage } from './utils/pagination/getTotalPage';
 
 const items = getNumbers(1, 42).map(n => `Item ${n}`);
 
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,25 +1,52 @@
...
 const items = getNumbers(1, 42).map(n => `Item ${n}`);
 
 export const App: React.FC = () => {
+  const [perPage, setPerPage] = useState<EPaginationSize>(
+    EPaginationSize.Medium,
+  );
+
+  const [currentPage, setCurrentPage] = useState(1);
+
+  const perPageSizeHandler = (e: ChangeEvent<HTMLSelectElement>) => {
```
BudnikOleksii, line 18:
```suggestion
  const handlePerPageChange = (event: ChangeEvent<HTMLSelectElement>) => {
```
- Avoid one-letter variable names
- handler name should start from `handle` and if it's a prop from `on` https://javascript.plainenglish.io/handy-naming-conventions-for-event-handler-functions-props-in-react-fc1cbb791364

src/App.tsx
```diff
@@ -1,25 +1,52 @@
...
+    const pageSize = +e.target.value;
+
+    if (pageSize === perPage) return;
```
BudnikOleksii, line 21:
```suggestion
    if (pageSize === perPage) {
      return;
    }
```

src/App.tsx
```diff
@@ -1,25 +1,52 @@
...
+
+    setPerPage(pageSize);
+    setCurrentPage(1);
+  };
+
+  const onPageChange = (page: number) => setCurrentPage(+page);
```
BudnikOleksii, line 27: It's already a number so you don't need unary plus. And this function looks redundant at all

BudnikOleksii (approved):
LGTM 🚀

PR #1110, opened 2024-11-29
my solution

minimal005:
[DEMO LINK](https://minimal005.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,17 +1,51 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
+import { Pagination } from './components/Pagination';
+import cn from 'classnames';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42).map(n => `Item ${n}`);
+const items: string[] = getNumbers(1, 42).map(n => `Item ${n}`);
+
+interface Props {
+  currentPage: number;
+  perPage: number;
+}
+const visibleItems = (props: Props) => {
+  const { currentPage, perPage } = props;
+  const start = currentPage * perPage - perPage;
+  const end = currentPage * perPage;
+
+  return items.slice(start, end);
+};
 
 export const App: React.FC = () => {
+  const [total] = useState<string[]>(items);
+  const [visiblePages, setVisiblePages] = useState<number>(9);
+  const [perPage, setPerPage] = useState<number>(5);
+  const [currentPage, setCurrentPage] = useState<number>(1);
+
+  const totalPages: number[] = getNumbers(1, Math.ceil(total.length / perPage));
+  const showPages = visibleItems({ currentPage, perPage });
+  const onPageChange = (page: number) => {
+    setCurrentPage(page);
+  };
+
+  const onNextPage = (sign: '+' | '-') => {
+    if (sign === '+') {
+      setCurrentPage(currentPage + 1);
+    } else {
+      setCurrentPage(currentPage - 1);
+    }
+  };
+
   return (
     <div className="container">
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,25 +1,78 @@
...
+import { Pagination } from './components/Pagination';
+import cn from 'classnames';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-const items = getNumbers(1, 42).map(n => `Item ${n}`);
+interface Item {
+  id: number;
+  title: string;
+}
+const items: Item[] = getNumbers(1, 42).map(n => ({
```
VolodymyrKirichenko: what the `n`? Change it

src/App.tsx
```diff
@@ -1,25 +1,78 @@
...
+
+  return items.slice(start, end);
+};
 
 export const App: React.FC = () => {
+  const [total] = useState<Item[]>(items);
+  const [visiblePages, setVisiblePages] = useState<number>(9);
+  const [perPage, setPerPage] = useState<number>(5);
+  const [currentPage, setCurrentPage] = useState<number>(1);
+
```
VolodymyrKirichenko: u don't need it If u pass a simple type as an initial value - the type is determined automatically(number, string, boolean, etc..)

```suggestion
  const [visiblePages, setVisiblePages] = useState(9);
  const [perPage, setPerPage] = useState(5);
  const [currentPage, setCurrentPage] = useState(1);

```

src/App.tsx
```diff
@@ -1,25 +1,78 @@
...
   return (
     <div className="container">
       <h1>Items with Pagination</h1>
 
       <p className="lead" data-cy="info">
-        Page 1 (items 1 - 5 of 42)
+        Page {currentPage} (items {(currentPage - 1) * perPage + 1} -{' '}
+        {visiblePages === currentPage ? total.length : currentPage * perPage} of{' '}
+        {items.length})
```
VolodymyrKirichenko: better to use interpolation here

src/App.tsx
```diff
@@ -1,25 +1,78 @@
...
-            className="form-control">
+            className="form-control"
+            value={perPage}
+            onChange={event => {
+              setPerPage(+event.target.value);
+              setCurrentPage(1);
+              setVisiblePages(() =>
+                Math.ceil(items.length / +event.target.value),
+              );
+            }}
```
VolodymyrKirichenko: create handle function for it

src/App.tsx
```diff
@@ -32,78 +85,53 @@ export const App: React.FC = () => {
...
-      {/* Move this markup to Pagination */}
       <ul className="pagination">
-        <li className="page-item disabled">
+        <li className={cn('page-item', { disabled: currentPage === 1 })}>
           <a
             data-cy="prevLink"
             className="page-link"
             href="#prev"
-            aria-disabled="true">
+            aria-disabled={currentPage === 1 ? 'true' : 'false'}
```
VolodymyrKirichenko: `===` always returns a Boolean type. Will be the same. Use it for all such cases
```suggestion
            aria-disabled={currentPage === 1}
```

src/App.tsx
```diff
@@ -1,25 +1,78 @@
-import React from 'react';
+/* eslint-disable max-len */
```
VolodymyrKirichenko: don't use disable for the whole file

VolodymyrKirichenko (changes requested):
Good job, but let's make some fixes 🥹

src/App.tsx
StasSokolov1: still not fixed

StasSokolov1 (approved):
Looks good to me, just one minor comment is still not fixed

PR #1194, opened 2025-03-10
Solution

KateShepel:
[DEMO LINK](https://KateShepel.github.io/react_pagination/)


src/App.tsx
```diff
@@ -1,110 +1,44 @@
...
-        </li>
-      </ul>
-      <ul>
-        <li data-cy="item">Item 1</li>
-        <li data-cy="item">Item 2</li>
-        <li data-cy="item">Item 3</li>
-        <li data-cy="item">Item 4</li>
-        <li data-cy="item">Item 5</li>
-      </ul>
+      <ItemList itemFrom={itemFrom} itemTo={itemTo} />
```
alexander-ignatow, line 41: It's a minor thing, but the API of `ItemList` looks a bit inconsistent

You are using `currentPage` and `perPage` - and that is enough to calculate the tems, displayed in the list

If you need, you may still calculate 
```
  const itemFrom = currentPage * perPage - perPage + 1;
  const itemTo = Math.min(itemFrom + perPage - 1, items.length);
```

But do this inside ItemList component

alexander-ignatow:
```suggestion
      <ItemList perPage={perPage} currentPage={currentPage} />
```

alexander-ignatow (approved):
Good job, but please, see my comments for the future

PR #1243, opened 2025-04-22
feat: add task solution

s1rserg:
[DEMO LINK](https://s1rserg.github.io/react_pagination/)


s1rserg:
@mateacademy-ai-mentor 
1. and 2, I cannot recalculate current page because i wont pass tests. It needs to be equal to one after change on perpage selector.
```
PerPage selelector

      √ should allow to select any per page from the list

      √ should set 1st page on change
```

2.

PR #1275, opened 2025-05-31
Solution

TetianaHodis1308:
[DEMO LINK](https://TetianaHodis1308.github.io/react_pagination/)


meryam99 (approved):
Nice job!

src/App.tsx
```diff
@@ -1,109 +1,69 @@
...
+import { Pagination } from './components/Pagination';
 
 export const App: React.FC = () => {
+  const [value, setValue] = useState(5);
+
+  const [currentPage, setCurrentPage] = useState(1);
+
+  const start = currentPage * value - value + 1;
+  const end = start + value - 1 < 42 ? start + value - 1 : 42;
+  const total = 42;
```
BudnikOleksii: `total` you can move outside the component and it looks like a constant. Also, use it instead of magic values in the `end` varaible

src/App.tsx
```diff
@@ -1,109 +1,69 @@
...
+  // const items = getNumbers(start, total).map(n => `Item ${n}`);
```
BudnikOleksii: Remove all comments

BudnikOleksii (approved):
Good work 👍

src/App.tsx
TetianaHodis1308: I've already done it, but I'm not sure if it's right.

PR #1332, opened 2025-08-13
add pagination

YuliaProkipchuk:
[DEMO LINK](https://YuliaProkipchuk.github.io/react_pagination/)


MikeKorotych (approved):
looks and works very nice 🤘

YuliaProkipchuk:
@mateacademy-ai-mentor in this case perPage will always be greater than 0 because of the selection options and default value is also set for perPage

PR #1345, opened 2025-08-25
Added final solution

YevheniiKa:
[DEMO LINK]
https://YevheniiKa.github.io/react_pagination/

### DIFF
--- a/src/components/Pagination/Pagination.tsx
+++ b/src/components/Pagination/Pagination.tsx
@@ -3,7 +3,7 @@ import { getNumbers } from '../../utils';
 import cn from 'classnames';
 type Props = {
   total: number;
-  perPage: string;
+  perPage: number;
   currentPage: number;
   onPageChange: (page: number) => void;
 };
@@ -21,16 +21,13 @@ export const Pagination: React.FC<Props> = ({
   function getItemsPerPage() {
     const copyItems = items;
 
-    return copyItems.slice(
-      (currentPage - 1) * +perPage,
-      currentPage * +perPage,
-    );
+    return copyItems.slice((currentPage - 1) * perPage, currentPage * perPage);
   }
 
   const visibleItems = getItemsPerPage();
   const pages: number[] = [];
 
-  for (let i = 1; i <= Math.ceil(total / +perPage); i++) {
+  for (let i = 1; i <= Math.ceil(total / perPage); i++) {
     pages.push(i);
   }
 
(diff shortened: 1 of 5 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,98 @@
...
+  return (
+    <>
+      <ul className="pagination">
+        <li className={cn('page-item', { disabled: currentPage === 1 })}>
+          <a
+            data-cy="prevLink"
+            className="page-link"
+            href="#prev"
+            aria-disabled={currentPage === 1}
+            onClick={() => onPageChange(pages.indexOf(currentPage))}
```
DidZhara: Using indexOf here seems wrong and chooses wrong page. Instead of going to 2 you will pass it. Similar issue in another one onClick

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,98 @@
...
+          className={cn('page-item', {
+            disabled: currentPage === pages[pages.length - 1],
+          })}
+        >
+          <a
+            aria-disabled={currentPage === pages[pages.length - 1]}
+            data-cy="nextLink"
+            className="page-link"
+            href="#next"
+            onClick={() => onPageChange(pages.indexOf(currentPage) + 2)}
```
DidZhara: Without event.preventDefault() there could be bugs with rendering

DidZhara (changes requested):

src/components/Pagination/Pagination.tsx
YevheniiKa: > Without event.preventDefault() there could be bugs with rendering

Hey mate, you check not needed commit, all implements already added in [fe2aaaf](https://github.com/mate-academy/react_pagination/pull/1345/commits/fe2aaafcf1e59698867e7a6bcde8e168f79b6f10)

src/components/Pagination/Pagination.tsx
YevheniiKa: > Using indexOf here seems wrong and chooses wrong page. Instead of going to 2 you will pass it. Similar issue in another one onClick

Hey mate, you check not needed commit, all implements already added in [fe2aaaf](https://github.com/mate-academy/react_pagination/pull/1345/commits/fe2aaafcf1e59698867e7a6bcde8e168f79b6f10)

PR #1372, opened 2025-10-12
Develop

DmytroLymar:
[DEMO LINK](https://DmytroLymar.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -16,10 +16,14 @@ function spliceItems(
 }
 
 export const App: React.FC = () => {
+  const DEFAULT_ITEMS_PER_PAGE = 5;
+  const DEFAULT_PAGE = 1;
+
   const [searchParams, setSearchParams] = useSearchParams();
 
-  const itemsPerPage = Number(searchParams.get('perPage')) || 5;
-  const currentPage = Number(searchParams.get('page')) || 1;
+  const itemsPerPage =
+    Number(searchParams.get('perPage')) || DEFAULT_ITEMS_PER_PAGE;
+  const currentPage = Number(searchParams.get('page')) || DEFAULT_PAGE;
 
   const total = items.length;
   const start = (currentPage - 1) * itemsPerPage;
@@ -27,7 +31,7 @@ export const App: React.FC = () => {
 
   const splicedItems = spliceItems(items, start, end);
 
-  const onItemsPerPageChange = (
+  const HandleItemsPerPageChange = (
     event: React.ChangeEvent<HTMLSelectElement>,
   ) => {
     const perPage = Number(event.target.value);
@@ -35,7 +39,7 @@ export const App: React.FC = () => {
     setSearchParams({ page: '1', perPage: String(perPage) });
   };
 
-  const onPageChange = (page: number) => {
+  const handlePageChange = (page: number) => {
     setSearchParams({ page: String(page), perPage: String(itemsPerPage) });
   };
 
@@ -54,7 +58,7 @@ export const App: React.FC = () => {
             id="perPageSelector"
             className="form-control"
             value={itemsPerPage}
-            onChange={onItemsPerPageChange}
+            onChange={HandleItemsPerPageChange}
           >
             <option value="3">3</option>
             <option value="5">5</option>
@@ -72,7 +76,7 @@ export const App: React.FC = () => {
         total={total}
         itemsPerPage={itemsPerPage}
         currentPage={currentPage}
-        onPageChange={onPageChange}
+        onPageChange={handlePageChange}
       />
       <ul>
         {splicedItems.map(el => (
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -31,7 +31,7 @@ export const App: React.FC = () => {
 
   const splicedItems = spliceItems(items, start, end);
 
-  const HandleItemsPerPageChange = (
+  const handleItemsPerPageChange = (
     event: React.ChangeEvent<HTMLSelectElement>,
   ) => {
     const perPage = Number(event.target.value);
@@ -58,7 +58,7 @@ export const App: React.FC = () => {
             id="perPageSelector"
             className="form-control"
             value={itemsPerPage}
-            onChange={HandleItemsPerPageChange}
+            onChange={handleItemsPerPageChange}
           >
             <option value="3">3</option>
             <option value="5">5</option>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,8 @@
-import React, { useState } from 'react';
+import React from 'react';
 import './App.css';
 import { getNumbers } from './utils';
 import { Pagination } from './components/Pagination';
+import { useSearchParams } from 'react-router-dom';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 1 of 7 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,79 @@
...
+  currentPage = 1,
+  onPageChange,
+}) => {
+  const pageCount = Math.max(1, Math.ceil(total / itemsPerPage));
+  const canNext = currentPage < pageCount;
+  const canPrev = currentPage > 1;
+  const pages = Array.from({ length: pageCount }, (_, index) => index + 1);
+
+  const onSelectPage = (page: number) => {
+    if (page !== currentPage && page >= 1 && page <= pageCount) {
```
781calmstate: you use a lot of `1`. some magic number that only you know why should be `1`.

let's make it more clear for other developers and create a const with a descriptive name for it

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,79 @@
...
+  itemsPerPage,
+  currentPage = 1,
+  onPageChange,
+}) => {
+  const pageCount = Math.max(1, Math.ceil(total / itemsPerPage));
+  const canNext = currentPage < pageCount;
+  const canPrev = currentPage > 1;
+  const pages = Array.from({ length: pageCount }, (_, index) => index + 1);
+
+  const onSelectPage = (page: number) => {
```
781calmstate: prefix `on` before the handler name is a bit wrong here.
the prefix `on` is used for props, the prefix `handle` is used for handlers you create inside of component.

So if you have the component which accepts the select page callback, it should look like this:


```suggestion
  const handleSelectPage = (page: number) => {...}
  
  return (
    <Component onSelectPage={handleSelectPage} />
  )
```

more about naming in react you can find [here](https://dev.to/kristiyanvelkov/react-js-naming-convention-lcg)

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,79 @@
...
+      <li className={cn('page-item', { disabled: !canPrev })}>
+        <a
+          data-cy="prevLink"
+          className="page-link"
+          href="#prev"
+          aria-disabled={!canPrev}
+          onClick={event => {
+            event.preventDefault();
+            onSelectPage(currentPage - 1);
+          }}
```
781calmstate: if you have inline callback that takes more than one line - it's probably good idea to move it into separated handler function

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,79 @@
...
+              onClick={event => {
+                event.preventDefault();
+                onSelectPage(page);
+              }}
+            >
+              {page}
+            </a>
+          </li>
+        );
+      })}
```
781calmstate, line 61: not a mistake, but could be component - something like `PaginationItem`

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,79 @@
...
+
+        return (
+          <li key={page} className={cn('page-item', { active: isCurrent })}>
+            <a
+              data-cy="pageLink"
+              className="page-link"
+              href={`#${page}`}
+              onClick={event => {
+                event.preventDefault();
+                onSelectPage(page);
```
781calmstate: same as above

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,79 @@
...
+      })}
+      <li className={cn('page-item', { disabled: !canNext })}>
+        <a
+          data-cy="nextLink"
+          className="page-link"
+          href="#next"
+          aria-disabled={!canNext}
+          onClick={event => {
+            event.preventDefault();
+            onSelectPage(currentPage + 1);
```
781calmstate: same as above about callbacks

src/App.tsx
```diff
@@ -1,25 +1,61 @@
...
+  end: number,
+): string[] {
+  return itemsList.slice(start, end);
+}
+
 export const App: React.FC = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+
+  const itemsPerPage = Number(searchParams.get('perPage')) || 5;
+  const currentPage = Number(searchParams.get('page')) || 1;
```
781calmstate: 5, 1 - magic numbers, their purpose is unclear for others

src/App.tsx
```diff
@@ -1,25 +1,61 @@
...
+
+  const total = items.length;
+  const start = (currentPage - 1) * itemsPerPage;
+  const end = Math.min(total, currentPage * itemsPerPage);
+
+  const splicedItems = spliceItems(items, start, end);
+
+  const onItemsPerPageChange = (
```
781calmstate: same about naming

src/App.tsx
```diff
@@ -1,25 +1,61 @@
...
+    event: React.ChangeEvent<HTMLSelectElement>,
+  ) => {
+    const perPage = Number(event.target.value);
+
+    setSearchParams({ page: '1', perPage: String(perPage) });
+  };
+
+  const onPageChange = (page: number) => {
```
781calmstate: same about naming

781calmstate (changes requested):

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,77 @@
...
+};
+
+export const Pagination: React.FC<Props> = ({
+  total,
+  itemsPerPage,
+  currentPage = 1,
+  onPageChange,
+}) => {
+  const FIRST_PAGE = 1;
+  const MIN_PAGE_COUNT = 1;
```
781calmstate, line 19: those are static values, you should move them out of component's body

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,77 @@
...
+  total: number;
+  itemsPerPage: number;
+  currentPage?: number;
+  onPageChange: (page: number) => void;
+};
+
+export const Pagination: React.FC<Props> = ({
+  total,
+  itemsPerPage,
+  currentPage = 1,
```
781calmstate, line 15: instead of one I guess you should use the `FIRST_PAGE`

src/App.tsx
```diff
@@ -1,25 +1,65 @@
...
+  itemsList: string[],
+  start: number,
+  end: number,
+): string[] {
+  return itemsList.slice(start, end);
+}
+
 export const App: React.FC = () => {
+  const DEFAULT_ITEMS_PER_PAGE = 5;
+  const DEFAULT_PAGE = 1;
```
781calmstate, line 20: same as above about static consts

src/App.tsx
```diff
@@ -1,25 +1,65 @@
...
+  const end = Math.min(total, currentPage * itemsPerPage);
+
+  const splicedItems = spliceItems(items, start, end);
+
+  const handleItemsPerPageChange = (
+    event: React.ChangeEvent<HTMLSelectElement>,
+  ) => {
+    const perPage = Number(event.target.value);
+
+    setSearchParams({ page: '1', perPage: String(perPage) });
```
781calmstate, line 39: shouldn't it be `String(FIRST_PAGE)`

781calmstate (approved):
Good job! Even though there is still a room for improvement :)

PR #1380, opened 2025-10-29
solution

aCTORr8:
[DEMO LINK](https://aCTORr8.github.io/react_pagination/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React, {useState} from 'react';
+import React, { useState } from 'react';
 import './App.css';
 import { getNumbers } from './utils';
 import { Pagination } from './components/Pagination';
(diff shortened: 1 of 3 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,91 @@
...
+            }}
+          >
+            {page}
+          </a>
+        </li>
+      ))}
+      <li
+        className={
+          currentPage === totalPages ? 'page-item disabled' : 'page-item'
+        }
```
Anton-Kuchmasov, line 77: Best practice is using [classnames ](https://www.npmjs.com/package/classnames)lib in such cases

src/App.tsx
```diff
@@ -1,25 +1,42 @@
...
+  const [currentPage, setCurrentPage] = useState<number>(1);
+
+  const total = items.length;
+  const start = (currentPage - 1) * perPage;
+  const end = start + perPage;
+
+  const handlePerPageSelector = (
+    event: React.ChangeEvent<HTMLSelectElement>,
+  ) => {
+    setPerPage(+event.target.value);
```
Anton-Kuchmasov, line 19: If `evetn.target.value` will be `undefined`, it will be 
`setPerPage(NaN)`

Consider check, if `event.target.value`, or just return from function

Anton-Kuchmasov (approved):
Well done!

Check out comments below - they might be helpful

PR #1453, opened 2026-03-09
Added solution for pagination

MykolaFatkullin:


MykolaFatkullin:
@mateacademy-ai-mentor setting key using index it's the best approach so we don't need to add items into function

PR #1466, opened 2026-03-26
Implement pagination + React Router*

Dianaeva:
[DEMO LINK](https://dianaeva.github.io/react_pagination/)

### DIFF
--- a/src/components/Pagination/Pagination.tsx
+++ b/src/components/Pagination/Pagination.tsx
@@ -1,1 +1,65 @@
-export const Pagination = () => {};
+import React from 'react';
+import cn from 'classnames';
+
+import { PageItem } from './PageItem';
+
+type Props = {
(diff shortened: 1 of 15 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,65 @@
...
+
+      {[...Array(totalPages)].map((_, i) => {
+        const page = i + 1;
+
+        return (
+          <PageItem
+            title={page.toString()}
+            dataCy="pageLink"
+            href={`#${page}`}
+            key={page}
```
polosanya, line 43: It is not good practice to use index as key

polosanya (changes requested):
Demo link doesn't work, please fix it
Image

polosanya (approved):

PR #1526, opened 2026-07-08
add solution

m-berezovskyy:


src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,92 @@
...
+          data-cy="prevLink"
+          className="page-link"
+          href="#/"
+          aria-disabled={currentPage === 1}
+          onClick={e => {
+            if (currentPage > 1) {
+              handlePageClick(currentPage - 1, e);
+            } else {
+              e.preventDefault();
+            }
```
2pasha: move to separate fn

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,92 @@
...
+          className="page-link"
+          href="#/"
+          aria-disabled={currentPage === totalPages}
+          onClick={e => {
+            if (currentPage < totalPages) {
+              handlePageClick(currentPage + 1, e);
+            } else {
+              e.preventDefault();
+            }
+          }}
```
2pasha: move to separate fn

2pasha (changes requested):
good job, but you should pass all CI tests

Image

also please take a look on comments below

2pasha (changes requested):
⬆️

m-berezovskyy:
Hi @2pasha, thanks for the feedback! I've extracted the onClick logic into separate functions as you suggested.

I also have a quick question about the CI tests. Before my recent commits (adding React Router), my local `npm test` ran perfectly without any errors, but the CI pipeline showed 25 failing tests. 

Could you please clarify why there is such a difference between my local environment and the CI? Is there a specific command (like Cypress e2e) I should run locally to catch these specific errors before pushing my code? 

Thanks!

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,96 @@
...
+    }
+  };
+
+  if (totalPages <= 1) {
+    return null;
+  }
+
+  return (
+    <ul className="pagination" data-cy="pagination">
+      <li className={`page-item ${currentPage === 1 ? 'disabled' : ''}`}>
```
Denys-Kravchuk9988: It's better to use `classNames` here

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,96 @@
...
+          onClick={handlePrev}
+        >
+          «
+        </a>
+      </li>
+
+      {pages.map(page => (
+        <li
+          key={page}
+          className={`page-item ${page === currentPage ? 'active' : ''}`}
```
Denys-Kravchuk9988: It's better to use `classNames` here

src/components/Pagination/Pagination.tsx
```diff
@@ -1 +1,96 @@
...
+            href="#/"
+            onClick={e => handlePageClick(page, e)}
+          >
+            {page}
+          </a>
+        </li>
+      ))}
+
+      <li
+        className={`page-item ${currentPage === totalPages ? 'disabled' : ''}`}
```
Denys-Kravchuk9988: It's better to use `classNames` here

Denys-Kravchuk9988 (changes requested):
Good job!

A few things to improve:

2pasha (approved):
lgtm 🔥

about tests: probably because system difference, because tests are running on Linux and you probably using another OS. But for now it's not a problem, because your solution is fine

